PR #4, opened 2019-03-28
Improve and support HA deployment mode

liusheng:
This change Improve labkeeper to support deploy latest openlab environment. And also this support HA deployment mode.

partially implement: # theopenlab/openlab#217


etc/nodepool/allinone-nodepool.yaml.j2
```diff
@@ -14,47 +14,39 @@ zookeeper-servers:
 
 labels:
   - name: ubuntu-xenial
-    min-ready: 1
+    min-ready: 0
+  - name: ubuntu-xenial-gpu
```
ZhengZhenyu, line 18: are these change necessary？

inventory/allinone-ha.yaml
```diff
@@ -0,0 +1,160 @@
+all:
+  hosts:
+    bastion:
+      ansible_connection: local
+    allinone01:
+      ansible_host: 192.168.211.20
+    allinone02:
+      ansible_host: 192.168.211.174
+    zk03:
+      ansible_host: 192.168.211.151
```
ZhengZhenyu, line 10: Could we make these vars so user can pass?

etc/unison/openlab-log-ha.prf.j2
```diff
@@ -0,0 +1,13 @@
...
+batch = true
+owner = true
+group = true
+perms = -1
+repeat = 1800
+retry = 3
+sshargs = -i /var/lib/zuul/.ssh/id_rsa -C
+fastcheck = true
+log = true
+logfile = /var/log/unison/unison.log
```
ZhengZhenyu, line 13: could be better to have comments on what these config means, it starts with # as line1

etc/nodepool/allinone-nodepool.yaml.j2
liusheng: not necessary, but according to now the allinone, just like a backup by the way :)

etc/unison/openlab-log-ha.prf.j2
liusheng: OK, will do

inventory/allinone-ha.yaml
liusheng: this yaml file just like a config file for different deployment mode. and I have mentioned in README to check these before deployment. not only these IP addresses, there are some other options in this file also need to check before deployment.

sure yes, we can make these vars allow passing, but I think maybe it is can be implement in a wrapping script e.g. `deploy.sh` in future improvement.

etc/nodepool/allinone-nodepool.yaml.j2
bzhaoopenstack: We must make sure the min-ready not change in Online openlab env to maintain the configuration not change on both side.

inventory/allinone-ha.yaml
```diff
@@ -0,0 +1,167 @@
...
+        nodepool-builder:
+          children:
+            nodepool-builder-master:
+              hosts: allinone01
+            nodepool-builder-slave:
+              hosts: allinone02
+              vars:
+                service_state: stopped
+          vars:
+            nodepool_builder_id: "8fcf0492-b972-45cc-828f-efd6b301b46e"
```
bzhaoopenstack, line 26: The nodepool builder id is also configured manually?

playbooks/keepalived.yaml
```diff
@@ -0,0 +1,86 @@
...
+      copy:
+        src: "{{ labkeeper_src_dir }}/openlab.pem"
+        dest: /etc/keepalived/openlab.pem
+        mode: 0600
+
+    - name: Enable ip_vs module
+      shell: |
+        modprobe ip_vs
+        modprobe ip_vs_wrr
+        lsmod |grep ip_vs
```
bzhaoopenstack, line 31: lsmod |grep ip_vs, this line is unecessary, but a check to make sure the ip_vs is loaded already.

playbooks/keepalived.yaml
```diff
@@ -0,0 +1,86 @@
...
+      args:
+        executable: /bin/bash
+
+- name: Config keepalived master
+  hosts: keepalived-master
+  become: yes
+  tasks:
+    - name: Get interface of default route
+      shell: ip route |grep default | awk '{print $NF}'
```
bzhaoopenstack: Why I run this cmd on my env, it return me a "onlink", it's not a valid interface name.
It shoud be awk '{print $(NF-1)}'

playbooks/keepalived.yaml
```diff
@@ -0,0 +1,86 @@
...
+
+    - name: Start keepalived service
+      service: name=keepalived state=started
+
+- name: Config keepalived slave
+  hosts: keepalived-slave
+  become: yes
+  tasks:
+    - name: Get interface of default route
+      shell: ip route |grep default | awk '{print $NF}'
```
bzhaoopenstack: NF-1

inventory/allinone-ha.yaml
liusheng: yes, I was told by zhenyu that we need to keep same nodepool builder id both  `master` and `slave`

playbooks/keepalived.yaml
liusheng: yes, that is my purpose ：）

inventory/allinone-ha.yaml
wangxiyuan: ++

wangxiyuan (commented):
LGMT overall, On more thing which can be done in a follow-up:

the `ssh_host_ecdsa_key` pair should be the same between master zuul node and slave zuul node as well. It can be hard-coded getting from current openlab production env.

PR #36, opened 2019-04-15
Add update_service func

wangxiyuan:
1. Add update_service func which will be used by keepalived. Note that this is not for operators so that we won't expose cli.

2. Fixed the bug that service's updated_at is always None.

3. Fixed the bug that the service's node role is always the same.

4. Fixed the bug that `node set` cli doesn't work.

5. Add `node_name_filter` and `node_role_filter` for `list_services` function

6. Expose `node_name` for service cmds.

### DIFF
--- a/openlabcmd/openlabcmd/cli.py
+++ b/openlabcmd/openlabcmd/cli.py
@@ -117,22 +117,30 @@ class OpenLabCmd(object):
 
     def _add_ha_service_cmd(self, parser):
         # openlab ha service
-        cmd_ha_service = parser.add_parser('service', help='Manage HA service.')
+        cmd_ha_service = parser.add_parser('service',
+                                           help='Manage HA service.')
         cmd_ha_service_subparsers = cmd_ha_service.add_subparsers(
             title='service', dest='service')
         # openlab ha service list
         cmd_ha_service_list = cmd_ha_service_subparsers.add_parser(
             'list', help='List all services.')
         cmd_ha_service_list.set_defaults(func=self.ha_service_list)
+        cmd_ha_service_list.add_argument(
+            '--node', action='append',
+            help='Filter the services with the specified node name.')
+        cmd_ha_service_list.add_argument(
+            '--role', action='append',
+            choices=['master', 'slave', 'zookeeper'],
+            help='Filter the services with the specified node role.')
         # openlab ha service get
         cmd_ha_service_get = cmd_ha_service_subparsers.add_parser(
-            'get', help='Get a services.')
+            'get', help='Get a service.')
         cmd_ha_service_get.set_defaults(func=self.ha_service_get)
         cmd_ha_service_get.add_argument('name', help='service name.')
         cmd_ha_service_get.add_argument(
             '--role', choices=['master', 'slave', 'zookeeper'],
             help="The role of the node where the service run. It must be "
-                 "sepcified if the service is 'mysql' or 'zookeeper'.")
+                 "specified if the service is in %s." % service.MIXED_SERVICE)
 
     def _add_ha_cmd(self, parser):
         # openlab ha
@@ -272,7 +280,7 @@ class OpenLabCmd(object):
 
     @_zk_wrapper
     def ha_service_list(self):
-        result = self.zk.list_services()
+        result = self.zk.list_services(self.args.node, self.args.role)
         if self.args.format == 'pretty':
             print(utils.format_output('service', result))
         else:
@@ -283,7 +291,8 @@ class OpenLabCmd(object):
         if (self.args.name.lower() in service.MIXED_SERVICE and
                 not self.args.role):
             raise argparse.ArgumentTypeError(
-                'Role must be specified if service is mysql or zookeeper.')
+                'Role must be specified if service is in '
+                '%s.' % service.MIXED_SERVICE)
         result = self.zk.get_service(self.args.name.lower(), self.args.role)
         if self.args.format == 'pretty':
             print(utils.format_output('service', result))
--- a/openlabcmd/openlabcmd/node.py
+++ b/openlabcmd/openlabcmd/node.py
@@ -10,17 +10,14 @@ class NodeStatus(object):
 
 class Node(object):
     def __init__(self, name, role, type, ip, heartbeat=None, alarmed=None,
-                 status=None):
+                 status=None, **kwargs):
         self.name = name
         self.role = role
         self.type = type
         self.ip = ip
-        if not heartbeat:
-            self.heartbeat = 0
-        if not alarmed:
-            self.alarmed = False
-        if not status:
-            self.status = NodeStatus.INITIALIZING
+        self.heartbeat = heartbeat or 0
+        self.alarmed = alarmed or False
+        self.status = status or NodeStatus.INITIALIZING
 
     def to_zk_data(self):
         node_dict = {
@@ -34,7 +31,6 @@ class Node(object):
         }
         return json.dumps(node_dict).encode('utf8')
 
-    @classmethod
-    def from_dict(cls, name, role, n_type, ip, heartbeat=None, alarmed=None,
-                  status=None):
-        return cls(name, role, n_type, ip, heartbeat, alarmed, status)
+    @ classmethod
+    def from_dict(cls, d):
+        return cls(**d)
--- a/openlabcmd/openlabcmd/service.py
+++ b/openlabcmd/openlabcmd/service.py
@@ -26,27 +26,31 @@ class ServiceStatus(object):
     INITIALIZING = 'initializing'
     UP = 'up'
     DOWN = 'down'
-    Restarting = 'restarting'
+    RESTARTING = 'restarting'
+
+    @property
+    def all_status(self):
+        return [self.INITIALIZING, self.UP, self.DOWN, self.RESTARTING]
 
 
 class Service(object):
-    def __init__(self, name, node_role):
+    def __init__(self, name, node_role, alarmed=None, alarmed_at=None,
+                 restarted=None, restarted_at=None, is_necessary=None,
+                 status=None, **kwargs):
         self.name = name
-        self.alarmed = False
-        self.alarmed_at = None
-        self.updated_at = None
-        self.restarted = False
-        self.restarted_at = None
-        self.is_necessary = None
         self.node_role = node_role
-        self.status = ServiceStatus.INITIALIZING
+        self.alarmed = alarmed or False
+        self.alarmed_at = alarmed_at
+        self.restarted = restarted or False
+        self.restarted_at = restarted_at
+        self.is_necessary = is_necessary
+        self.status = status or ServiceStatus.INITIALIZING
 
     def to_zk_data(self):
         node_dict = {
             'name': self.name,
             'alarmed': self.alarmed,
             'alarmed_at': self.alarmed_at,
-            'updated_at': self.updated_at,
             'restarted': self.restarted,
             'restarted_at': self.restarted_at,
             'is_necessary': self.is_necessary,
@@ -54,6 +58,10 @@ class Service(object):
             'status': self.status
         }
         return json.dumps(node_dict).encode('utf8')
+
+    @ classmethod
+    def from_dict(cls, d):
+        return cls(**d)
 
 
 class NecessaryService(Service):
--- a/openlabcmd/openlabcmd/zk.py
+++ b/openlabcmd/openlabcmd/zk.py
@@ -35,8 +35,22 @@ class ZooKeeper(object):
                                          "object.")
         self._last_retry_log = 0
 
-    def _bytesToDict(self, data):
-        return json.loads(data.decode('utf8'))
+    def _bytesToDict(self, node_data, include_created_at=False,
+                     include_updated_at=False):
+        """ Convert zookeeper node data into dict."""
+        node_dict = node_data[0]
+        # ctime and mtime is millisecond in zk.
+        ctime, mtime = node_data[1].ctime, node_data[1].mtime
+
+        return_info = json.loads(node_dict.decode('utf8'))
+        if include_created_at:
+            return_info['created_at'] = datetime.datetime.fromtimestamp(
+                ctime / 1000).isoformat()
+        if include_updated_at:
+            return_info['updated_at'] = datetime.datetime.fromtimestamp(
+                mtime / 1000).isoformat()
+
+        return return_info
 
     def _connection_listener(self, state):
         if state == KazooState.LOST:
@@ -117,19 +131,15 @@ class ZooKeeper(object):
                 nodes.append(node_info)
         except kze.NoNodeError:
             return []
-        return sorted(nodes)
+        return sorted(nodes, key=lambda x: x['name'])
 
     @_client_check_wrapper
     def get_node(self, node_name):
         try:
             node_info = self.client.get('/ha/%s' % node_name)
-            # ctime and mtime is millisecond in zk.
-            ctime, mtime = node_info[1].ctime, node_info[1].mtime
-            node_info = self._bytesToDict(node_info[0])
-            node_info['created_at'] = datetime.datetime.fromtimestamp(
-                ctime/1000).isoformat()
-            node_info['updated_at'] = datetime.datetime.fromtimestamp(
-                mtime/1000).isoformat()
+
+            node_info = self._bytesToDict(node_info, include_created_at=True,
+                                          include_updated_at=True)
             return node_info
         except kze.NoNodeError:
             raise exceptions.ClientError('Node %s not found.' % node_name)
@@ -162,7 +172,8 @@ class ZooKeeper(object):
                                  service_type == 'necessary' else
                                  service.UnnecessaryService)
                 for service_name in service_names:
-                    new_service = service_class(service_name, role)
+                    new_service = service_class(service_name,
+                                                node_role.split('_')[0])
                     self.client.create(
                         new_service_path + '/%s' % service_name,
                         value=new_service.to_zk_data())
@@ -193,7 +204,7 @@ class ZooKeeper(object):
             node_info['role'] = role
 
         node_info.update(kwargs)
-        new_node = node.Node.from_dict(**node_info)
+        new_node = node.Node.from_dict(node_info)
         self.client.set(path, value=new_node.to_zk_data())
 
         node_info = self.get_node(node_name)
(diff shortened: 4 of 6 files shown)


openlabcmd/openlabcmd/zk.py
```diff
@@ -206,37 +207,100 @@ def delete_node(self, node_name):
...
             for service_name in self.client.get_children(path):
                 service_path = path + '/' + service_name
                 service_info = self.client.get(service_path)
+                mtime = service_info[1].mtime
                 service_info = self._bytesToDict(service_info[0])
+                service_info['updated_at'] = datetime.datetime.fromtimestamp(
+                    mtime / 1000).isoformat()
+                service_info['node_name'] = exist_node['name']
                 result.append(service_info)
         return sorted(result)
```
kiwik: Sort a zookeeper object list, so what is expected return value? order by `node_name`?

openlabcmd/openlabcmd/zk.py
```diff
@@ -206,37 +207,100 @@ def delete_node(self, node_name):
...
-                exceptions.ClientError("Can't find service %s" % service_name)
+    def _get_service_role(self, service_name):
+        if (service_name in
+                service.service_mapping['master_service']['necessary'] +
+                service.service_mapping['master_service']['unnecessary']):
+            return 'master'
+        elif (service_name in
+              service.service_mapping['slave_service']['necessary'] +
+              service.service_mapping['slave_service']['unnecessary']):
+            return 'slave'
```
kiwik, line 274: Please add some comments about this logic, I originally thought I should add the following comments until I found the logic in cli.py :)

> we can't use service_name to decide service role, like: `mysql` and `zookeeper`, they are both in master and slave nodes. Imagine we input `mysql` into the function, always return `master`

openlabcmd/openlabcmd/zk.py
```diff
@@ -206,37 +207,100 @@ def delete_node(self, node_name):
...
+
+        exceptions.ClientError("Can't find service %s" % service_name)
```
kiwik: Raise?

openlabcmd/openlabcmd/zk.py
```diff
@@ -206,37 +207,114 @@ def delete_node(self, node_name):
...
+            return 'slave'
 
+        exceptions.ClientError("Can't find service %s" % service_name)
+
+    def _get_service_path_and_node(self, service_name, role):
         for exist_node in self.list_nodes():
             if exist_node['role'] == role:
                 path = '/ha/%s/%s_service/%s' % (exist_node['name'], role,
                                                  service_name)
-                result = self.client.get(path)
```
kiwik, line 239: Why do not check the service existed? if service_name=`not_existed_service`?

openlabcmd/openlabcmd/zk.py
```diff
@@ -206,37 +207,114 @@ def delete_node(self, node_name):
...
+    def get_service(self, service_name, role=None):
+        if not role:
+            role = self._get_service_role(service_name)
+        path, srv_node = self._get_service_path_and_node(service_name, role)
+        result = self.client.get(path)
+        mtime = result[1].mtime
+        result = self._bytesToDict(result[0])
+        result['updated_at'] = datetime.datetime.fromtimestamp(
+            mtime / 1000).isoformat()
+        result['node_name'] = srv_node['name']
```
kiwik, line 293: I see line#277~#281 logic multiple times in different places, please add a function to handle this.

openlabcmd/openlabcmd/service.py
```diff
@@ -26,27 +26,31 @@ class ServiceStatus(object):
     INITIALIZING = 'initializing'
     UP = 'up'
     DOWN = 'down'
-    Restarting = 'restarting'
+    RESTARTING = 'restarting'
+
+    @property
+    def all_services(self):
```
kiwik: s/all_services/all_status/

openlabcmd/openlabcmd/zk.py
```diff
@@ -206,37 +207,114 @@ def delete_node(self, node_name):
...
+                raise exceptions.ValidationError('restarted should be '
+                                                 'boolean.')
+            old_service['restarted'] = restarted
+            old_service['restarted_at'] = current_time
+        if status:
+            if status not in service.ServiceStatus.all_services:
+                raise exceptions.ValidationError(
+                    'status should be in %s.' %
+                    service.ServiceStatus.all_services)
+            old_service['status'] = status
```
kiwik, line 326: Don't forget to update `updated_at` attribute.

openlabcmd/openlabcmd/zk.py
```diff
@@ -206,37 +207,114 @@ def delete_node(self, node_name):
...
+        if not role:
+            if service_name in service.MIXED_SERVICE:
+                raise exceptions.ValidationError(
+                    "Role must be specified if service is in "
+                    "%s." % service.MIXED_SERVICE)
+            role = self._get_service_role(service_name)
+        path, _ = self._get_service_path_and_node(service_name, role)
+        current_time = datetime.datetime.now().isoformat()
+
+        if alarmed:
```
kiwik: How to update `alarmed` to false?

openlabcmd/openlabcmd/zk.py
```diff
@@ -206,37 +207,114 @@ def delete_node(self, node_name):
...
+            if not isinstance(alarmed, bool):
+                raise exceptions.ValidationError('alarmed should be boolean.')
+            old_service['alarmed'] = alarmed
+            old_service['alarmed_at'] = current_time
+        if restarted:
```
kiwik: ditto

kiwik (changes requested):

openlabcmd/openlabcmd/zk.py
wangxiyuan: `updated_at` is handled by zookeeper automatically. Every node in zookeeper has some build-in properties like ctime (created_at), mtime(updated_at), once the node info is changed, `mtime` will be changed as well.

openlabcmd/openlabcmd/zk.py
wangxiyuan: oh, my fault here. It should be `if alarmed is not None`

openlabcmd/openlabcmd/zk.py
wangxiyuan: I'll merge them into `_bytesToDict` func

openlabcmd/openlabcmd/zk.py
wangxiyuan: This function is only used for generate a path which will be used later, if the service doesn't exist, it'll raise NotFound error later.

openlabcmd/openlabcmd/zk.py
wangxiyuan: ++

openlabcmd/openlabcmd/zk.py
wangxiyuan: OK. will do

openlabcmd/openlabcmd/zk.py
wangxiyuan: Done

openlabcmd/openlabcmd/service.py
wangxiyuan: Done

openlabcmd/openlabcmd/zk.py
kiwik: OK, thanks for your clarification

openlabcmd/openlabcmd/zk.py
kiwik: I see the column `updated_at` https://github.com/theopenlab/labkeeper/pull/36/files#diff-fe16cf954dd7bc0d309b1d1d1f2e1441R39 , that will be mapped to `mtime` automatically?

wangxiyuan: Yes, see L50, I mapped `mtime` to `updated_at`  there when fetching data from zk.

openlabcmd/openlabcmd/cli.py
```diff
@@ -124,6 +124,11 @@ def _add_ha_service_cmd(self, parser):
...
             'list', help='List all services.')
         cmd_ha_service_list.set_defaults(func=self.ha_service_list)
+        cmd_ha_service_list.add_argument(
+            '--node', help='Filter the services with the specified node.')
+        cmd_ha_service_list.add_argument(
+            '--role', choices=['master', 'slave', 'zookeeper'],
+            help='Filter the services with the specified node role.')
         # openlab ha service get
         cmd_ha_service_get = cmd_ha_service_subparsers.add_parser(
             'get', help='Get a services.')
```
huangtianhua: Not relate with this pr: Get a service?

wangxiyuan: Please take a look at the pr commit message. I did a lot change in this pr.

openlabcmd/openlabcmd/node.py
```diff
@@ -10,17 +10,14 @@ class NodeStatus(object):
...
         self.ip = ip
-        if not heartbeat:
-            self.heartbeat = 0
-        if not alarmed:
-            self.alarmed = False
-        if not status:
-            self.status = NodeStatus.INITIALIZING
+        self.heartbeat = heartbeat if heartbeat else 0
+        self.alarmed = alarmed if alarmed else False
+        self.status = status if status else NodeStatus.INITIALIZING
```
huangtianhua: Why not set the default values for these parameters, then we can set them simply?

openlabcmd/openlabcmd/service.py
```diff
@@ -26,27 +26,31 @@ class ServiceStatus(object):
...
-        self.restarted_at = None
-        self.is_necessary = None
         self.node_role = node_role
-        self.status = ServiceStatus.INITIALIZING
+        self.alarmed = alarmed if alarmed else False
+        self.alarmed_at = alarmed_at if alarmed_at else None
+        self.restarted = restarted if restarted else False
+        self.restarted_at = restarted_at if restarted_at else None
+        self.is_necessary = is_necessary if is_necessary else None
+        self.status = status if status else ServiceStatus.INITIALIZING
```
huangtianhua: same.

openlabcmd/openlabcmd/node.py
wangxiyuan: It works here.

But it is not a good custom for python developer IMO. It works terribly for inherit class or dict/list args. Here is an article explain clearly: https://graysonkoonce.com/always-use-none-for-default-args-in-python/

I'm used to using `None` for default value.

openlabcmd/openlabcmd/node.py
```diff
@@ -10,17 +10,14 @@ class NodeStatus(object):
...
         self.role = role
         self.type = type
         self.ip = ip
-        if not heartbeat:
-            self.heartbeat = 0
-        if not alarmed:
-            self.alarmed = False
-        if not status:
-            self.status = NodeStatus.INITIALIZING
+        self.heartbeat = heartbeat if heartbeat else 0
```
liusheng: `= heartbeat or 0`

liusheng: can the `heartbeat` be initialized with 0 ? that may cause `if heartbeat` clause to be `False`

openlabcmd/openlabcmd/node.py
```diff
@@ -10,17 +10,14 @@ class NodeStatus(object):
...
+        self.alarmed = alarmed if alarmed else False
```
liusheng: alarmed or False

openlabcmd/openlabcmd/node.py
```diff
@@ -34,7 +31,6 @@ def to_zk_data(self):
         }
         return json.dumps(node_dict).encode('utf8')
 
-    @classmethod
-    def from_dict(cls, name, role, n_type, ip, heartbeat=None, alarmed=None,
-                  status=None):
-        return cls(name, role, n_type, ip, heartbeat, alarmed, status)
+    @ classmethod
+    def from_dict(cls, kwargs):
```
liusheng: not sure if `kwargs` is a good  input variable name. generally, it is used to be refer to `input keyword arguments`, but here you need to input a dict, right ? not similar to the above old function. I am also not very sure.

openlabcmd/openlabcmd/service.py
```diff
@@ -26,27 +26,31 @@ class ServiceStatus(object):
...
-        self.restarted = False
-        self.restarted_at = None
-        self.is_necessary = None
         self.node_role = node_role
-        self.status = ServiceStatus.INITIALIZING
+        self.alarmed = alarmed if alarmed else False
+        self.alarmed_at = alarmed_at if alarmed_at else None
+        self.restarted = restarted if restarted else False
+        self.restarted_at = restarted_at if restarted_at else None
+        self.is_necessary = is_necessary if is_necessary else None
```
liusheng: why need `... if ... else...` ? these vars already have the default `None` value

openlabcmd/openlabcmd/zk.py
```diff
@@ -35,8 +35,22 @@ def __init__(self, config=None):
                                          "object.")
         self._last_retry_log = 0
 
-    def _bytesToDict(self, data):
-        return json.loads(data.decode('utf8'))
+    def _bytesToDict(self, node_data, include_created_at=False,
+                     include_updated_at=False):
+        """ Convert zookeeper node data into dict."""
+        node_dict = node_data[0]
```
liusheng, line 41: node_data is a list and length always is 1 ?

openlabcmd/openlabcmd/zk.py
```diff
@@ -206,37 +217,116 @@ def delete_node(self, node_name):
         self.client.delete(path, recursive=True)
 
     @_client_check_wrapper
-    def list_services(self):
+    def list_services(self, node_name_filter=None, node_role_filter=None):
+        """
+        List the services in the HA deployment.
+        :param node_name_filter: The node filter.
+        :type node_name_filter: list or string.
```
liusheng, line 224: does the `--role` and `--node` arguments can be parsed to a list currently ?

openlabcmd/openlabcmd/zk.py
```diff
@@ -206,37 +217,116 @@ def delete_node(self, node_name):
...
+                raise exceptions.ValidationError(
+                    "Role must be specified if service is in "
+                    "%s." % service.MIXED_SERVICE)
+            role = self._get_service_role(service_name)
+        path, _ = self._get_service_path_and_node(service_name, role)
+        current_time = datetime.datetime.now().isoformat()
+
+        if alarmed is not None:
+            if not isinstance(alarmed, bool):
+                raise exceptions.ValidationError('alarmed should be boolean.')
```
liusheng: boolean value

openlabcmd/openlabcmd/cli.py
huangtianhua: I mean this should be 'Get a service'.

openlabcmd/openlabcmd/service.py
huangtianhua: +1

openlabcmd/openlabcmd/cli.py
```diff
@@ -124,6 +124,11 @@ def _add_ha_service_cmd(self, parser):
         cmd_ha_service_list = cmd_ha_service_subparsers.add_parser(
             'list', help='List all services.')
         cmd_ha_service_list.set_defaults(func=self.ha_service_list)
+        cmd_ha_service_list.add_argument(
+            '--node', help='Filter the services with the specified node.')
```
huangtianhua: "...... specified node name" is more clear.

openlabcmd/openlabcmd/zk.py
```diff
@@ -206,37 +217,116 @@ def delete_node(self, node_name):
...
+            if node_role_filter and exist_node['role'] not in node_role_filter:
+                continue
             path = '/ha/%s/%s_service' % (exist_node['name'],
                                           exist_node['role'])
             for service_name in self.client.get_children(path):
                 service_path = path + '/' + service_name
                 service_info = self.client.get(service_path)
-                service_info = self._bytesToDict(service_info[0])
+                service_info = self._bytesToDict(service_info,
+                                                 include_updated_at=True)
```
huangtianhua, line 254: Just a question: when the update_at is changed for service, there is not set or update operate for service, right? if it's changed, why created_at is not needed here?

huangtianhua: sorry, if it's can be changed.....

openlabcmd/openlabcmd/node.py
wangxiyuan: Oh, `kwargs` is not a suitable name here. I'll change it.

openlabcmd/openlabcmd/zk.py
wangxiyuan: node_data  is a list, length is always 2.

openlabcmd/openlabcmd/zk.py
huangtianhua: sorry, misunderstand, ignore this.

openlabcmd/openlabcmd/node.py
wangxiyuan: the type of heartbeat  is int. `heartbeat=0` means the service is just initialized.

openlabcmd/openlabcmd/zk.py
wangxiyuan: I can't see any improvement except simpler input format?

openlabcmd/openlabcmd/zk.py
```diff
@@ -35,8 +35,22 @@ def __init__(self, config=None):
...
+        node_dict = node_data[0]
+        # ctime and mtime is millisecond in zk.
+        ctime, mtime = node_data[1].ctime, node_data[1].mtime
+
+        return_info = json.loads(node_dict.decode('utf8'))
+        if include_created_at:
+            return_info['created_at'] = datetime.datetime.fromtimestamp(
+                ctime / 1000).isoformat()
+        if include_updated_at:
+            return_info['updated_at'] = datetime.datetime.fromtimestamp(
```
kiwik, line 50: Got it, thank you @wangxiyuan

kiwik (approved):
LGTM

openlabcmd/openlabcmd/zk.py
liusheng: got, thx

openlabcmd/openlabcmd/zk.py
liusheng: may you can add `action='append', default=[]` to the arguments

openlabcmd/openlabcmd/zk.py
```diff
@@ -206,37 +217,117 @@ def delete_node(self, node_name):
...
+            role = self._get_service_role(service_name)
+        path, srv_node = self._get_service_path_and_node(service_name, role)
+        result = self.client.get(path)
+        result = self._bytesToDict(result, include_updated_at=True)
+        result['node_name'] = srv_node['name']
+        return result
+
+    @_client_check_wrapper
+    def update_service(self, service_name, role=None, alarmed=None,
+                       restarted=None, status=None, **kwargs):
```
liusheng, line 298: didn't see any usage of the `**kwargs` argument in this function

wangxiyuan: See L328

openlabcmd/openlabcmd/zk.py
wangxiyuan: I don't think merge different filters into once arg is a good way. It's not comfortable for user experience IMO. Take `openstack server list` for example, each filter use the arg itself.

openlabcmd/openlabcmd/zk.py
liusheng: oh, sorry.

openlabcmd/openlabcmd/zk.py
wangxiyuan: emm, I think i misunderstand your message before.  See new PS please

liusheng: maybe you don't need L229~L240 if you added `default=[]` in the arguments

wangxiyuan: openlabcmd is not only used for operator, but also for keepalived or some other manage tools in the future.  L229-L240 can make the input more flexible.

liusheng: OK,

liusheng (approved):

PR #40, opened 2019-04-18
Update keepalived script and refactor related part in labkeeper

bzhaoopenstack:
For this loop.
The following tests are passed.
1. unecessary serivice down, post an issue, if the issue is timeout, do switch.
2. necessary service down, post an issue, do switch.
3. master host down/ keepalived is down(heartbeat fail to update, then timeout), post an issue, do switch.


playbooks/keepalived.yaml
```diff
@@ -45,62 +35,34 @@
...
+        zk_cli_conf: "{{ labkeeper_config_git_dest }}/openlabcmd/etc/openlab.conf"
+        heartbeat_internal: 100
+        ori_master_ip: "{{ hostvars[groups['zuul-web-master'][0]].ansible_host }}"
+        ori_backup_ip: "{{ hostvars[groups['zuul-web-slave'][0]].ansible_host }}"
+        slave_ip: "{{ hostvars[groups['zuul-web-slave'][0]].ansible_host }}"
+        slave_backup_ip: "{{ hostvars[groups['zuul-web-slave'][0]].ansible_host }}"
 
     - name: install script dependency
       pip:
-        name: ['PyGithub', 'argparse', 'requests', 'html', 'datetime']
```
liusheng, line 103: please use a list of packages for `pip`, the with_items of pip installation usage has been deprecated.

etc/keepalived/kp_local_loop.py
```diff
@@ -0,0 +1,581 @@
...
+
+# Heartbeat Interanl time(seconds)
+HEARTBEAT_INTERNAL = {{ heartbeat_internal }}
+
+# Simpledns provider
+base_api_url = 'https://api.dnsimple.com/v2/'
+DOMAIN_NAME = 'openlabtesting.org'
+DOMAIN_NAME1 = 'status.openlabtesting.org'
+DOMAIN_NAME2 = 'logs.openlabtesting.org'
+DOMAIN_NAME_BK = 'logs-bak.openlabtesting.org'
```
wangxiyuan, line 32: what's this used for?

etc/keepalived/kp_local_loop.py
```diff
@@ -0,0 +1,581 @@
...
+            zk_cli.update_node(self.name, alarmed=True)
+            self.alarmed = True
+
+
+def get_zk_cli():
+    global GLOBAL_ZK
+    if not GLOBAL_ZK:
+        cfg = configparser.ConfigParser().read(ZK_CLI_CONF)
+        zk_cli = zk.ZooKeeper(cfg)
+        zk_cli.connect()
```
wangxiyuan, line 253: you must disconnect zk as well after every time you connect to zk.

etc/keepalived/kp_local_loop.py
```diff
@@ -0,0 +1,581 @@
...
+            Service(service['name'], service['alarmed'],
+                    service['alarmed_at'], service['restarted'],
+                    service['restarted_at'], service['is_necessary'],
+                    service['node_tole'], service['status']))
+        if service['is_necessary']:
+            local_node.necessary_service_names.append(service['name'])
+
+    global GLOBAL_NODE
+    GLOBAL_NODE = local_node
+    return GLOBAL_NODE
```
wangxiyuan: no need to set GLOBAL_NODE here, just return `local_node`

etc/keepalived/kp_local_loop.py
```diff
@@ -0,0 +1,581 @@
...
+    zk_cli = get_zk_cli()
+    node_name = get_local_hostname()
+    zk_cli.connect()
+    zk_node = zk_cli.get_node(node_name)
+    local_node = Node(zk_node['name'], zk_node['role'],
+                      zk_node['type'], zk_node['ip'],
+                      zk_node['heartbeat'], zk_node['alarmed'],
+                      zk_node['status'])
+
+    for service in zk_cli.list_services(node=node_name):
```
wangxiyuan: `list_services(node_name_filter=node_name)`

etc/keepalived/kp_local_loop.py
```diff
@@ -0,0 +1,581 @@
...
+        self.status = run_systemctl_command(SYSTEMCTL_STATUS, self.name)
+        zk_cli = get_zk_cli()
+        zk_cli.update_service(self.name, role=self.node_role,
+                              status=self.status)
+        return self.status != 'up'
+
+    def is_alarmed_timeout(self):
+        over_time = parse_isotime(
+            self.alarmed_at) + datetime.timedelta(
+            days=2)
```
wangxiyuan, line 182: it's good to be configurable for the `days`, hour is better. We can save it in zk

wangxiyuan (changes requested):

etc/keepalived/kp_local_loop.py
bzhaoopenstack: Used for change DNS

etc/keepalived/kp_local_loop.py
bzhaoopenstack: Right, We can support it by storing necessary info into zk.

etc/keepalived/kp_local_loop.py
bzhaoopenstack: Why? I prefer to use a global single session during access z, like we use zkCli.sh..I test it, it's OK to do this..  Or is there any other issues I need to fix if don't close the connection?

etc/keepalived/kp_local_loop.py
bzhaoopenstack: Thanks, I realize the interface had been updated.

etc/keepalived/kp_local_loop.py
bzhaoopenstack: ;-)

etc/keepalived/kp_local_loop.py
wangxiyuan: If you disconnect zk by hand in the end of the script, zk will log info that `session is disconnected`.

If you don't, when the script is done, zk will log warning that `session is lost` which is not a correct way.

etc/keepalived/kp_local_loop.py
wangxiyuan: just return local_node here and then set `GLOBAL_NODE=load_from_zk()` at L275, so that you done't need declare `GLOBAL_NODE` once again here.

etc/keepalived/kp_local_loop.py
```diff
@@ -0,0 +1,665 @@
...
+        zk_cli = zk.ZooKeeper(cfg)
+        zk_cli.connect()
+        GLOBAL_ZK = zk_cli
+    return GLOBAL_ZK
+
+
+def load_from_zk():
+    zk_cli = get_zk_cli()
+    node_name = get_local_hostname()
+    zk_cli.connect()
```
wangxiyuan: you have connected to zk at L248, why connect it again?

etc/keepalived/kp_local_loop.py
```diff
@@ -0,0 +1,665 @@
...
+        cfg.read(ZK_CLI_CONF)
+        zk_cli = zk.ZooKeeper(cfg)
+        zk_cli.connect()
+        GLOBAL_ZK = zk_cli
+    return GLOBAL_ZK
+
+
+def load_from_zk():
+    zk_cli = get_zk_cli()
+    node_name = get_local_hostname()
```
wangxiyuan, line 260: I don't think this is worth to use a  new function.

etc/keepalived/kp_local_loop.py
```diff
@@ -0,0 +1,665 @@
...
+
+def local_node_service_process(node_obj):
+    for service_obj in node_obj.services:
+        treat_single_service(service_obj, node_obj)
+
+    node_obj.report_heart_beat()
+
+
+def treat_single_service(service_obj, node_obj):
+    if node_obj.is_maintained():
```
wangxiyuan: this can  be removed now. so as `is_maintained()`

wangxiyuan (changes requested):

etc/keepalived/kp_local_loop.py
```diff
@@ -0,0 +1,665 @@
...
+                self.name, self.node_role, self.node_type, alarmed=True)
+            self.alarmed = True
+            self.alarmed_at = updated_svc['alarmed_at']
+
+    def post_restarted_if_possible(self):
+        if not self.alarmed:
+            zk_cli = get_zk_cli()
+            updated_svc = zk_cli.update_service(
+                self.name, self.node_role, self.node_type,
+                restarted=True, status='restarting')
```
wangxiyuan, line 199: I didn't find the code to set 'restarting' to 'up' or 'down'?

etc/keepalived/kp_local_loop.py
```diff
@@ -0,0 +1,665 @@
...
+    node_obj.report_heart_beat()
+
+
+def treat_single_service(service_obj, node_obj):
+    if node_obj.is_maintained():
+        return
+
+    if service_obj.is_abnormal():
+        if service_obj.restarted:
+            if node_obj.is_necessary_service(service_obj.name):
```
wangxiyuan: service_obj has a property called `is_necessary`, why don't check it directly? I have to say that the list `necessary_service_names` in Service class is redundant

wangxiyuan (changes requested):

etc/keepalived/kp_local_loop.py
bzhaoopenstack: Please check in L145， each loop， if the current restart is different with restarting, it will store the new status in zk.

etc/keepalived/kp_local_loop.py
bzhaoopenstack: Ok, so you are agree with the single session, but need close the session at the end of script. Right?

etc/keepalived/kp_local_loop.py
bzhaoopenstack: Correct, remove it

etc/keepalived/kp_local_loop.py
bzhaoopenstack: Now, the node_name designation is different with openlabcmd format node name. So I just split this out for custmized the hostname to fit the gap.

etc/keepalived/kp_local_loop.py
bzhaoopenstack: Yeah, this code need to be refactor, this code is fit for the original openlabcmd code.

etc/keepalived/kp_local_loop.py
wangxiyuan: ++

etc/keepalived/kp_local_loop.py
wangxiyuan: Emm, right. So you **update** the service in `is_abnormal` function which I assume is a function for **check**. It's a little strange.

etc/keepalived/kp_local_loop.py
bzhaoopenstack: done

etc/keepalived/kp_local_loop.py
bzhaoopenstack: done

etc/keepalived/kp_local_loop.py
bzhaoopenstack: done

etc/keepalived/kp_local_loop.py
```diff
@@ -0,0 +1,693 @@
...
+    script_load_pre_check(local_node)
+    if local_node.status in ['maintaining', 'down']:
+        return
+    local_node_service_process(local_node)
+    if local_node.role != 'zookeeper':
+        oppo_node_process()
+
+    # At the end of the script, we'd better to close zk session.
+    zk_cli = get_zk_cli()
+    zk_cli.close()
```
wangxiyuan: `zk_cli.disconnect()`

etc/keepalived/kp_local_loop.py
```diff
@@ -0,0 +1,693 @@
...
+                seconds=self.hb_interval_time)
+        except ValueError:
+            # The heartbeat is not formatted, this must be the kp on the node
+            # is not finish the first loop. We just return False, as we are in
+            # the initializing process.
+            return False
+        current_time = datetime.datetime.utcnow().replace(tzinfo=iso8601.UTC)
+        return current_time > over_time
+
+    def is_necessary_service(self, service_name):
```
wangxiyuan: this function can be removed

etc/keepalived/kp_local_loop.py
```diff
@@ -0,0 +1,693 @@
...
+def shut_down_all_services(node_obj):
+    for service in node_obj.services:
+        run_systemctl_command(SYSTEMCTL_STOP, service.name)
+    node_obj.services = []
+
+
+def setup_necessary_services_and_check(node_obj):
+    for service in node_obj.necessary_service_names:
+        run_systemctl_command(SYSTEMCTL_START, service)
+
```
wangxiyuan, line 377:
```
for service in node_obj.services:
    if service['is_necessary']:
        run_systemctl_command(SYSTEMCTL_START, service)
```

Then the `necessary_service_names` property can be removed totally.

wangxiyuan (changes requested):

etc/keepalived/kp_local_loop.py
bzhaoopenstack: done

etc/keepalived/kp_local_loop.py
bzhaoopenstack: done ;-)

etc/keepalived/kp_local_loop.py
bzhaoopenstack: done

etc/keepalived/kp_local_loop.py
bzhaoopenstack: I add a TODO on this function

wangxiyuan (approved):
LGTM, Let's merge it and take a fully test.

playbooks/keepalived.yaml
bzhaoopenstack: done

PR #49, opened 2019-04-23
openlabcmd py3 support

wangxiyuan:
1. Remove `commands` lib usage.
2. Add `enum` requirement only for py2.

### DIFF
--- a/openlabcmd/requirements.txt
+++ b/openlabcmd/requirements.txt
@@ -1,4 +1,5 @@
-pbr>=1.3
-PyYAML>=5.1
+enum;python_version=='2.7'
 kazoo
+pbr>=1.3
 prettytable
+PyYAML>=5.1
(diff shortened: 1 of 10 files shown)


openlabcmd/requirements.txt
```diff
@@ -1,4 +1,5 @@
-pbr>=1.3
-PyYAML>=5.1
+enum;python_version=='2.7'
```
kiwik, line 1: Question: The update only is supported by py3,  or both py2 and py3? I find the default python version is python 2.7 in OpenLab CI host (zuul and nodepool), so we need to make sure openlabcmd can work on python 2.7

wangxiyuan: openlabcmd support both py2 and py3.

`enum` in py3 is a built-in library, so we only install it in py2.

kiwik: OK, thank you

kiwik (approved):
LGTM

PR #56, opened 2019-04-25
Delete branch if config files not changed

huangtianhua:
To delete branch anyway after trying to sync config files updating.

Related-Bug: theopenlab/openlab#84


kiwik (approved):
Thank you.

PR #62, opened 2019-04-29
Refactor openlabcmd using object

wangxiyuan:
Use python `object` instead of python `dict`.

This will be used by keepalived easier.

### DIFF
--- a/openlabcmd/openlabcmd/zk.py
+++ b/openlabcmd/openlabcmd/zk.py
@@ -1,6 +1,5 @@
 import configparser
 import datetime
-import json
 import logging
 import time
 
@@ -34,23 +33,6 @@ class ZooKeeper(object):
             raise exceptions.ClientError("config should be a ConfigParser "
                                          "object.")
         self._last_retry_log = 0
-
-    def _bytesToDict(self, node_data, include_created_at=False,
-                     include_updated_at=False):
-        """ Convert zookeeper node data into dict."""
-        node_dict = node_data[0]
-        # ctime and mtime is millisecond in zk.
-        ctime, mtime = node_data[1].ctime, node_data[1].mtime
-
-        return_info = json.loads(node_dict.decode('utf8'))
-        if include_created_at:
-            return_info['created_at'] = datetime.datetime.fromtimestamp(
-                ctime / 1000).isoformat()
-        if include_updated_at:
-            return_info['updated_at'] = datetime.datetime.fromtimestamp(
-                mtime / 1000).isoformat()
-
-        return return_info
 
     def _connection_listener(self, state):
         if state == KazooState.LOST:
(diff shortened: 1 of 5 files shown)


openlabcmd/openlabcmd/zk.py
```diff
@@ -125,46 +107,44 @@ def wrapper(self, *args, **kwargs):
     def list_nodes(self):
         path = '/ha'
         try:
-            nodes = []
+            nodes_obj = []
```
ZhengZhenyu: tiny nit, could be better to name it node_objs

ZhengZhenyu (commented):
nice

openlabcmd/openlabcmd/zk.py
```diff
@@ -125,46 +107,44 @@ def wrapper(self, *args, **kwargs):
...
-            return node_info
+            node_bytes = self.client.get('/ha/%s' % node_name)
+            node_obj = node.Node.from_zk_bytes(node_bytes)
+            return node_obj
         except kze.NoNodeError:
             raise exceptions.ClientError('Node %s not found.' % node_name)
 
     @_client_check_wrapper
-    def create_node(self, name, role, n_type, ip):
+    def create_node(self, node_name, role, n_type, ip):
```
bzhaoopenstack: why not just "name"

openlabcmd/openlabcmd/zk.py
```diff
@@ -297,24 +271,23 @@ def update_service(self, service_name, role, n_type, alarmed=None,
             if not isinstance(alarmed, bool):
                 raise exceptions.ValidationError('alarmed should be boolean '
                                                  'value.')
-            old_service['alarmed'] = alarmed
-            old_service['alarmed_at'] = current_time
+            old_service.alarmed = alarmed
+            old_service.alarmed_at = current_time
```
bzhaoopenstack, line 277: I saw the current_time  using the local time zone. Right now the running host is UTC, so we just in the same time zone. How about use the same UTC time zone with specified UTC?

openlabcmd/openlabcmd/zk.py
```diff
@@ -297,24 +271,23 @@ def update_service(self, service_name, role, n_type, alarmed=None,
             if not isinstance(alarmed, bool):
                 raise exceptions.ValidationError('alarmed should be boolean '
                                                  'value.')
-            old_service['alarmed'] = alarmed
```
bzhaoopenstack, line 300: Also in here, we update a serivice using a node_type, but the definition of service obj doesn't contain it. How about to add it? as the node_name and node_role are there already and a single service can only match a role:type except the data plane services. Just a nit, as this function can also work

ZhengZhenyu:
Yes, checking again, are there any difference with ``name`` and ``node_name``?

openlabcmd/openlabcmd/zk.py
wangxiyuan: `updated_at` and `created_at` are generated by zookeeper internally, it uses local time. So that to keep the same with zookeeper, we should use local time as well.

openlabcmd/openlabcmd/zk.py
wangxiyuan: I just realized that `node_name` is unique. a service can be fetched by a command like `openlab ha service get <service_name> --node <node_name>` If so, both `node_role` and `node_type` can be removed from `service` obj.

Thought?

openlabcmd/openlabcmd/zk.py
bzhaoopenstack: emm, imaging that Cloud A using time zone A locally, Cloud B using time zone B locally, and Cloud C using time zone C. If we deploy zk cluster on those, which time zone it will use?  I;m afraid we may maintain multiple time zone format. Allright, let's forget this now, as it's not a real problem. We can leave it on our notes.

openlabcmd/openlabcmd/zk.py
bzhaoopenstack: Correct, let's make it unit function, not coupling so much.

bzhaoopenstack (approved):
gogogo

openlabcmd/openlabcmd/zk.py
wangxiyuan: This is a good point. ++ for UTC, since different cloud may live in different time zone. I'll fix it in the next pr. Thanks!

PR #64, opened 2019-04-30
Use UTC timezone

wangxiyuan:
Openlab HA nodes will be deployed in different cloud. Different cloud may in different timezone. We should keep the time the same inner the HA cluster.

### DIFF
--- a/openlabcmd/openlabcmd/node.py
+++ b/openlabcmd/openlabcmd/node.py
@@ -1,6 +1,8 @@
 import datetime
 import json
 
+import pytz
+
 
 class NodeStatus(object):
     INITIALIZING = 'initializing'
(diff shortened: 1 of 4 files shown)


bzhaoopenstack (approved):
lgtm

openlabcmd/openlabcmd/node.py
```diff
@@ -46,9 +48,9 @@ def update(self, update_dict):
     def from_zk_bytes(cls, zk_bytes):
         node_dict = json.loads(zk_bytes[0].decode('utf8'))
         # ctime and mtime is millisecond in zk.
-        ctime, mtime = zk_bytes[1].ctime, zk_bytes[1].mtime
+        ctime, mtime = zk_bytes[1].ctime / 1000, zk_bytes[1].mtime / 1000
```
liusheng, line 51: seems, you can use the `iso8601` to do this format converting, it is more convenient

wangxiyuan: I just convert the `int` type millisecond to second here . Can't see how iso8601 works here? Suggestion?

wangxiyuan:
recheck

PR #65, opened 2019-04-30
Refactor `service get` cmd

wangxiyuan:
The `node_name` is global unique in ha deployment. So that the service can be fetched with `node_name`, no need for role and type any more.


bzhaoopenstack (approved):
LGTM

PR #66, opened 2019-05-06
Remove keepalived

wangxiyuan:
Keepalived now is only used for timer tracker. This can be replaced by `systemd`. So that we don't need maintain keepalived itself.


playbooks/ha_service_install.yaml
```diff
@@ -55,5 +42,8 @@
       pip:
         name: ['six', 'iso8601', 'requests', 'configparser', 'PyGithub', 'openlabcmd']
```
huangtianhua: Seems we install openlabcmd  in playbook 'ha_openlabcmd_install.yaml'?

bzhaoopenstack:
er, just post a new PR, then I found this. how about focus on PR #67

PR #67, opened 2019-05-06
Split health checker from kp

bzhaoopenstack:
1. Split health checker from kp to systemd
2. Refact healthchecker to 3 parts, refresh, fix, switch.
3. Support Log function.
4. Fix a openlabcmd as the update function of Node obj, will failed to update heartbeat, as its default value is always 0.


etc/healthchecker/healthchecker.service
```diff
@@ -0,0 +1,5 @@
+[Unit]
+Description=Healthchecker
+
+[Service]
+ExecStart=/etc/healthchecker/healthchecker.sh
```
wangxiyuan: let's move to `/etc/openlab` folder to make things cleaner

etc/healthchecker/healthchecker.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+/etc/healthchecker/refresh.py
```
wangxiyuan: ditto

etc/healthchecker/healthchecker.timer
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=Run Openlab HA health checker scripts periodically
+
+[Timer]
+OnActiveSec=10s
+OnBootSec=300s
+OnStartupSec=10s
+OnUnitActiveSec=30s
```
wangxiyuan: 30s is to short IMO. How about 1-2 min

etc/healthchecker/refresh.py
```diff
@@ -0,0 +1,312 @@
+#!/usr/bin/python2
```
wangxiyuan: I'd like to see the tools can support py3 by default. Since the components in openlab are mainly running on py3.

etc/keepalived/kp_local_loop.py
```diff
@@ -16,8 +16,8 @@
 # ZK client config file
```
wangxiyuan: does this file still work? I assume it can be removed.

playbooks/healthchecker.yaml
```diff
@@ -0,0 +1,59 @@
...
+# License for the specific language governing permissions and limitations
+# under the License.
+---
+- name: Install healthchecker
+  hosts: healthchecker
+  become: yes
+  tasks:
+    - name: New a healthchecker directory
+      file:
+        path: /etc/healthchecker/
```
wangxiyuan: ditto, let's use `/etc/openlab`

playbooks/healthchecker.yaml
```diff
@@ -0,0 +1,59 @@
...
+      pip:
+        name: ['six', 'iso8601', 'requests', 'configparser', 'PyGithub', 'openlabcmd']
+
+    - name: config healthchecker
+      copy: src={{ item.src }} dest={{ item.dest }}
+      with_items:
+        - { src: "{{ labkeeper_config_git_dest }}/healthchecker/healthchecker.timer", dest: '/etc/systemd/system/healthchecker.timer'}
+        - { src: "{{ labkeeper_config_git_dest }}/healthchecker/healthchecker.service", dest: '/etc/systemd/system/healthchecker.service' }
+
+    # As systemd ansible module bugs not fix #36585 and #48786
```
wangxiyuan: Really? Have you tested it? I see 36585 is closed https://github.com/ansible/ansible/issues/36585 and 48786 has some fixed comment.

playbooks/site.yaml
```diff
@@ -31,7 +31,7 @@
 - import_playbook: ha_log_cfg_sync.yaml
 - import_playbook: ha_openlabcmd_install.yaml
 - import_playbook: apache.yaml
-- import_playbook: keepalived.yaml
+- import_playbook: healthchecker.yaml
```
wangxiyuan: let's rename to `ha_healthchecker` and move to L33 to keep the code cleaner.

wangxiyuan (changes requested):

etc/keepalived/kp_local_loop.py
bzhaoopenstack: No, I will remove it.

playbooks/healthchecker.yaml
bzhaoopenstack: Yeah, I have test it. If we use the systemd module like:
become: yes
systemd:
   name: healthchecker.service
   enabled: yes
   started: yes
Or any ways we can try to do this using systemd module, it will still give back an error "Could not find the requested service healthchecker.service: host", which like the issue mentioned, also those issues are not fixed in ansible, you can check the comments log..

playbooks/site.yaml
bzhaoopenstack: Done

etc/healthchecker/healthchecker.service
bzhaoopenstack: Done

etc/healthchecker/healthchecker.sh
bzhaoopenstack: Done

etc/healthchecker/healthchecker.timer
bzhaoopenstack: Depends on the sensitivity

etc/healthchecker/refresh.py
bzhaoopenstack: Yeah, all of them support py3

playbooks/healthchecker.yaml
bzhaoopenstack: Done

wangxiyuan (approved):

playbooks/healthchecker.yaml
wangxiyuan: Cool. good to know

PR #69, opened 2019-05-08
Fix openlabcmd 

wangxiyuan:
1. HA tools now are ruuning under py3 by default. Openlabcmd should
be under py3 as well.

2. When users set node from `mantained` to `up`, the heartbeat should
be refreshed as well.


bzhaoopenstack (commented):
good to go

ZhengZhenyu:
LGTM

Yikun:
LGTM

PR #70, opened 2019-05-09
refactor rsyncd to systemd

bzhaoopenstack:
1. refactor rsyncd on slave node to run based on systemd
2. Fix some nits in current playbooks

### DIFF
--- a/playbooks/ha_log_cfg_sync.yaml
+++ b/playbooks/ha_log_cfg_sync.yaml
@@ -25,15 +25,20 @@
         dest: /etc/rsyncd/rsyncd.conf
 
     - name: Start rsync daemon on slave nodes
+      template:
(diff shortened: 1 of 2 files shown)


wangxiyuan (approved):

kiwik (approved):
LGTM ,thank you.

playbooks/ha_log_cfg_sync.yaml
```diff
@@ -25,16 +25,21 @@
         dest: /etc/rsyncd/rsyncd.conf
 
     - name: Start rsync daemon on slave nodes
+      template:
+        src: "{{ labkeeper_config_git_dest }}/rsync/rsyncd.service"
```
liusheng, line 29: it is better to name the file to xxx.j2, even if this can also work, but looks werid.

PR #73, opened 2019-05-13
Refactor healthchecker

bzhaoopenstack:
make the hk distribulization.
Currently, All healthcheckers runs on each node are simple distributed.
Each HK will refresh local node status, but some specific process during oppo checks.
I test some scenarios with this scripts in simulated online ENV( 5 nodes in OTC).
1. Master node's necessary service dead.
    RES: try to restart the error service first, if its status is still down after restart. We will post an issue and do swith.
2. Master node's unecessary service dead.
   RES: try to restart the error service first, if its status is still down after restart. We just post an issue
3. Master node dead.
    1) 1 Master node dead.
        RES: This scenarios happens on the slave node which is oppo check. Slave node found master node can not ping and its heartbeat is overtime. Slave node will post an issue to report the master node can not accessed, then do the switch.
    2) 2 Master node dead.
        RES: This is nearly same with above, but this scenarios will post 2 issues. As currently the couple monitor to watch each other must be matched the same type , but different role. Like (master zuul <-> slave zuul)
4. Slave node's service dead.
    RES: currently, slave nodes just contains unecessary services. If one of them dead, try to restart first, if still down after restart, we just post an issue to report the service is dead.
5. Slave node dead.
   RES: whatever 1 or 2 slave nodes dead, we just post 1 or 2 issues.
6. zookeeper node service dead.
   RES: try to restart first, if still down after restart, we just post an issue to report the service is dead.


playbooks/ha_healthchecker.yaml
```diff
@@ -24,10 +24,7 @@
     - name: Copy necessary scripts
       template: src={{ item.src }} dest={{ item.dest }} mode=0755
       with_items:
-        - { src: "{{ labkeeper_config_git_dest }}/ha_healthchecker/ha_healthchecker.sh", dest: '/etc/openlab/ha_healthchecker/ha_healthchecker.sh'}
-        - { src: "{{ labkeeper_config_git_dest }}/ha_healthchecker/refresh.py", dest: '/etc/openlab/ha_healthchecker/refresh.py' }
-        - { src: "{{ labkeeper_config_git_dest }}/ha_healthchecker/fix.py", dest: '/etc/openlab/ha_healthchecker/fix.py'}
-        - { src: "{{ labkeeper_config_git_dest }}/ha_healthchecker/switch.py", dest: '/etc/openlab/ha_healthchecker/switch.py' }
+        - { src: "{{ labkeeper_config_git_dest }}/ha_healthchecker/ha_healthchecker.py", dest: '/etc/openlab/ha_healthchecker/ha_healthchecker.py' }
```
wangxiyuan: add .j2 format.

etc/ha_healthchecker/ha_healthchecker.py
```diff
@@ -0,0 +1,748 @@
...
+TARGET_CHANGE_IP = "{{ slave_ip }}"
+TARGET_CHANGE_BK_IP = "{{ slave_backup_ip }}"
+
+# General constants
+SYSTEMCTL_STATUS = 'status'
+SYSTEMCTL_RESTART = 'restart'
+SYSTEMCTL_STOP = 'stop'
+SYSTEMCTL_START = 'start'
+SYSTEMCTL_ENABLE = 'enable'
+SYSTEMCTL_DISABLE = 'disable'
```
wangxiyuan: I think these systemctl variable is useless. I can just use the strings

etc/ha_healthchecker/ha_healthchecker.py
```diff
@@ -0,0 +1,748 @@
...
+                update_dict['status'] = 'down'
+        if update_dict:
+            self.zk.update_service(service_obj.name, node_obj.name,
+                                   **update_dict)
+
+        if cur_status != 'up':
+            LOG.debug(
+                "Service %(name)s is abnormal. It's status is %(status)s.",
+                {'name': service_obj.name,
+                 'status': cur_status.upper()})
```
wangxiyuan: if the service is abnormal, it's always `down`, this error message should be rewrote.

etc/ha_healthchecker/ha_healthchecker.py
```diff
@@ -0,0 +1,748 @@
...
+        super(Refresher, self).__init__(zk, heartbeat_timeout)
+
+    def _local_node_service_process(self, node_obj):
+        service_objs = self.zk.list_services(node_name_filter=node_obj.name)
+        for service_obj in service_objs:
+            self._refresh_service(service_obj, node_obj)
+
+        self._report_heart_beat(node_obj)
+        if node_obj.alarmed and node_obj.role == 'slave':
+            self.zk.update_node(node_obj.name, alarmed=False)
```
wangxiyuan: let's move L151-L152 into L150. Once report heart beat, if `is_alarmed=True`, change it to `False` at the same time.

etc/ha_healthchecker/ha_healthchecker.py
```diff
@@ -0,0 +1,748 @@
...
+    def __init__(self, zk, heartbeat_timeout):
+        self.zk = zk
+        self.heartbeat_timeout = heartbeat_timeout
+        self.node = self.zk.get_node(socket.gethostname())
+        self.oppo_node = self._get_oppo_node()
+
+    def _get_oppo_node(self):
+        for zk_node in self.zk.list_nodes():
+            # During swtich, we can not get oppo node
+            if (zk_node.role != self.node.role and
```
wangxiyuan: It can be just: ` if zk_node.type == self.node.type and zk_node.name != self.node.name:`

etc/ha_healthchecker/ha_healthchecker.py
```diff
@@ -0,0 +1,748 @@
...
+
+    def run(self):
+        if self.node.status in ['maintaining', 'down']:
+            LOG.debug(
+                'Node %(name)s status is %(status)s, Skipping refresh.',
+                {'name': self.node.name,
+                 'status': self.node.status.upper()})
+            return
+        self._local_node_service_process(self.node)
+        self._oppo_node_check(self.oppo_node)
```
wangxiyuan:
```
if self.oppo_node:
    self._oppo_node_check(self.oppo_node)
```
then L190-L191 can be removed

etc/ha_healthchecker/ha_healthchecker.py
```diff
@@ -0,0 +1,748 @@
...
+                  {'name': node_obj.name, 'hb': hb})
+
+    def _oppo_node_check(self, oppo_node_obj):
+        if self.node.type == 'zookeeper':
+            return
+        if oppo_node_obj.status == 'maintaining':
+            return
+        if (not self._ping(oppo_node_obj.ip) and
+                self._is_check_heart_beat_overtime(oppo_node_obj)):
+            if oppo_node_obj.role == 'master' and oppo_node_obj.status == 'up':
```
wangxiyuan: no matter what role the oppo_node is, we should mark it to `down`,  right?

etc/ha_healthchecker/ha_healthchecker.py
```diff
@@ -0,0 +1,748 @@
...
+                    self._post_alarmed(oppo_node_obj)
+
+    def run(self):
+        if self.node.status in ['maintaining', 'down']:
+            LOG.debug('Node %(name)s status is %(status)s, Skipping fix.',
+                      {'name': self.node.name,
+                       'status': self.node.status.upper()})
+            return
+        self._local_node_service_process()
+        if self.node.role != 'zookeeper':
```
wangxiyuan: if self.oppo_node

etc/ha_healthchecker/ha_healthchecker.py
```diff
@@ -0,0 +1,748 @@
...
+class HealthChecker(object):
+    def __init__(self, config_file, heartbeat_timeout):
+        cfg = configparser.ConfigParser()
+        cfg.read(config_file)
+        self.zk = zk.ZooKeeper(cfg)
+        self.heartbeat_timeout = heartbeat_timeout
+
+    def run(self):
+        self.zk.connect()
+
```
wangxiyuan: I think all the check related to `maintaining` status can be moved to here. So that if the node is in `maintaining`, we can skip all the work below.

wangxiyuan (dismissed):

playbooks/ha_healthchecker.yaml
bzhaoopenstack: done

etc/ha_healthchecker/ha_healthchecker.py
bzhaoopenstack: I think it's not good to use string every where, like the status string in current code. But 'enable' and 'disable' can be removed for now.

etc/ha_healthchecker/ha_healthchecker.py
bzhaoopenstack: Yeah, the reason to do this is the first condition is the normal scenario, but the second is the during switch scenario, for make it clear, I post both condition here for review first. will remove in the end of commit

etc/ha_healthchecker/ha_healthchecker.py
bzhaoopenstack: Correct. Done

bzhaoopenstack: This logic is just to fix the down slave node manually, and re-online it.

etc/ha_healthchecker/ha_healthchecker.py
bzhaoopenstack: Correct, each loop will report this message till this node to be set to 'down', we just want a single message? If yes, we can introduce the restarted or alarmed here, or just log when svr.status == 'restarting'

etc/ha_healthchecker/ha_healthchecker.py
bzhaoopenstack: As we treat 'down' 'maintainning' for skip process. It's reasonable to set master node to down, as the host is done. We can use the down status to check whether we need to do the switch. For slave node, what we want to do is just report a issue, as we only care about master to slave switch. If before the switch we set the oppo slave node to down, we may skip so many process. Also we need a way to set it back, that why the L151 to L152, I only specify the alarmed slave node.

etc/ha_healthchecker/ha_healthchecker.py
bzhaoopenstack: 666, we may fix a zookeeper node bug.

etc/ha_healthchecker/ha_healthchecker.py
bzhaoopenstack: done

etc/ha_healthchecker/ha_healthchecker.py
bzhaoopenstack: Yeah, but we don't manage node obj in HealthChecker which is just a entry object.. So I think it's good to leave them in the workers objects.

bzhaoopenstack (commented):
ah, repo conflict, try to fix.......

etc/ha_healthchecker/ha_healthchecker.py.j2
```diff
@@ -0,0 +1,744 @@
...
+                update_dict['status'] = 'down'
+        if update_dict:
+            self.zk.update_service(service_obj.name, node_obj.name,
+                                   **update_dict)
+
+        if cur_status != 'up' and service_obj.status == 'restarting':
+            LOG.debug(
+                "Service %(name)s is abnormal. It's status is %(status)s.",
+                {'name': service_obj.name,
+                 'status': cur_status.upper()})
```
wangxiyuan: move the log to L164 is more clear. Then just : "Service %(name)s is down.", ({'name': service_obj.name})

etc/ha_healthchecker/ha_healthchecker.py.j2
```diff
@@ -0,0 +1,744 @@
...
+        super(Refresher, self).__init__(zk, heartbeat_timeout)
+
+    def _local_node_service_process(self, node_obj):
+        service_objs = self.zk.list_services(node_name_filter=node_obj.name)
+        for service_obj in service_objs:
+            self._refresh_service(service_obj, node_obj)
+
+        self._report_heart_beat(node_obj)
+        if node_obj.alarmed and node_obj.role == 'slave':
+            self.zk.update_node(node_obj.name, alarmed=False)
```
wangxiyuan: It's still here. Please merge it into L145

etc/ha_healthchecker/ha_healthchecker.py.j2
```diff
@@ -0,0 +1,744 @@
...
+        self._local_node_service_process()
+        if self.oppo_node:
+            self._oppo_node_check(self.oppo_node)
+
+
+class Switcher(Base):
+    def __init__(self, zk, heartbeat_timeout):
+        super(Switcher, self).__init__(zk, heartbeat_timeout)
+
+    def global_analysis(self):
```
wangxiyuan: this name can be just `def _is_need_switch`

wangxiyuan (changes requested):
LGTM overall. Just some nit. 

While the logic in Switcher looks a little redundance, we can improve it in the next PR.

etc/ha_healthchecker/ha_healthchecker.py.j2
bzhaoopenstack: O, forgot to remove

etc/ha_healthchecker/ha_healthchecker.py.j2
bzhaoopenstack: Done

etc/ha_healthchecker/ha_healthchecker.py.j2
bzhaoopenstack: done

wangxiyuan (approved):
LGTM

kiwik (approved):
LGTM

PR #81, opened 2019-05-15
Add exit errno in openlab check cmd

Yikun:
if the job check failed, will exit non-zero errno.

### DIFF
--- a/openlabcmd/openlabcmd/cli.py
+++ b/openlabcmd/openlabcmd/cli.py
@@ -203,6 +203,7 @@ class OpenLabCmd(object):
             plugins = base.Plugin.plugins
 
         cnt = len(cloud_list)
+        exit_flag = False
         for i, c in enumerate(cloud_list):
             header = "%s/%s. %s cloud check" % (i + 1, cnt, c)
             self._header_print(header)
@@ -214,6 +215,11 @@ class OpenLabCmd(object):
                 # the failed flag would be record when do check()
                 if self.args.recover and plugin.failed:
                     plugin.recover()
+                if plugin.failed:
+                    exit_flag = True
+
+        if exit_flag:
+            raise exceptions.ClientError("Error: cloud check failed.")
 
     def _zk_wrapper(func):
         def wrapper(self, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


bzhaoopenstack (approved):
lgtm

openlabcmd/openlabcmd/cli.py
```diff
@@ -214,6 +215,11 @@ def check(self):
                 # the failed flag would be record when do check()
                 if self.args.recover and plugin.failed:
                     plugin.recover()
+                if plugin.failed:
+                    exit_flag = 1
+
+        if not exit_flag:
+            exit(exit_flag)
```
wangxiyuan: Instead of `exit`, I suggest to raise ClientException error like other commands do to keep the same behavior.

wangxiyuan (changes requested):

openlabcmd/openlabcmd/cli.py
Yikun: Good idea, but in current implements, if we `raise ClientException`, the exit errno is still zero. because the "self.run" is not returned. We need change it to:
```python
    def _main(self):
        # ... ...
        return self.run()
```

Does it also have some impact on ha work? If not, I can address it in next PS.

[1] https://github.com/theopenlab/labkeeper/blob/0423574/openlabcmd/openlabcmd/cli.py#L339

wangxiyuan: do it!

Yikun: done

openlabcmd/openlabcmd/cli.py
```diff
@@ -329,10 +332,14 @@ def _initConfig(self):
...
-        self.parser = self.create_parser()
-        self.args = self.parser.parse_args()
-        self._initConfig()
-        self.run()
+        try:
+            self.parser = self.create_parser()
+            self.args = self.parser.parse_args()
+            self._initConfig()
+            return self.run()
+        except exceptions.ClientError as e:
```
wangxiyuan: We should catch all kinds of openlabcmd exception, not only ClientError , but also ValidationError. I think we can add a base Exception there and catch it.

wangxiyuan (changes requested):

openlabcmd/openlabcmd/cli.py
Yikun: done done done

huangtianhua (approved):
LGTM

wangxiyuan (approved):

PR #86, opened 2019-05-20
Fix alarmed status for node and support change switch_status operation.

bzhaoopenstack:
For now, there is no way to set node alarmed back to None, this patch
adds a way to self fix the alarmed status.
1. local services are OK. Set back to None, if node alarmed is True.
2. if the err services are all unecessary services and alarmed is not
timeout, also the local node is master.

Also, this patch introduces a way to set back the internal
switch_status.

Secondery, This patch support operator force setting the switch_status to start to
do a S/W swtich manually.

### DIFF
--- a/etc/ha_healthchecker/ha_healthchecker.py.j2
+++ b/etc/ha_healthchecker/ha_healthchecker.py.j2
@@ -714,15 +714,18 @@ class Switcher(Base):
                 'start' in set(res) and
                 'end' in set(res))
 
-    def _do_switch(self):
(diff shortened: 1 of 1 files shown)


wangxiyuan:
Please add `Related-Bug: theopenlab/openlab#218` in commit message.

wangxiyuan (approved):

wangxiyuan (approved):

kiwik (approved):
LGTM

PR #87, opened 2019-05-20
Update deploy script

wangxiyuan:
1. Merge some parameters to into one `action` to make cli clearner.
2. Fix a bug that `new-slave` action doesn't update master nodes.
3. Move HA deploy step to the last to ensure everything is OK before start health checker.
4. Separate mysql install and HA config steps to different file.

### DIFF
--- a/deploy.py
+++ b/deploy.py
@@ -13,22 +13,25 @@ def add_cli_args():
                                  "openlab-ha"],
                         help='OpenLab deployment type',
                         )
+    parser.add_argument('--action',
+                        choices=["deploy", "new-slave", "switch-role",
+                                 "show-graph", "show-ip"],
+                        default="deploy",
+                        help="The action that labkeeper supports. Default "
+                             "value is 'deploy'.\n"
+                             "'deploy': create a new OpenLab CI environment.\n"
+                             "'new-slave': create new slave nodes base on "
+                             "current master nodes.\n"
+                             "'switch-role': switch the master/slave role of "
+                             "current deployment. This will re-config the "
+                             "existing OpenLab environment.\n"
+                             "'show-graph': show the hosts graph of ansible "
+                             "inventory. It can be used with --with-vars to "
+                             "show more detail.\n"
+                             "'show-ip': show hosts ip address.\n")
     parser.add_argument('-u', '--user',
                         help='the Ansible remote user performing deployment, '
                              'default is "ubuntu" configured in ansible.cfg',
-                        )
-    parser.add_argument('--switch-role',
-                        action='store_true',
-                        help='switch the master/slave role of current deployment, '
-                             'this will re-config an existing OpenLab environment to '
-                             'switch the "master/slave" role of hosts',
-                        )
-    parser.add_argument('--new-slave',
-                        action='store_true',
-                        help='deploy the slave part of a CI cluster freshly, you need '
-                             'to specify the new slave hosts IP addresses with "--new-ip" '
-                             'argument, for example: '
-                             '--new-ip zuul02=192.168.5.5 --new-ip nodepool02=192.168.6.6',
                         )
     parser.add_argument('-e', '--extra-vars',
                         metavar="<key=value>",
@@ -36,30 +39,20 @@ def add_cli_args():
                         default=[],
                         help='extra vars passed to Ansible command',
                         )
-    parser.add_argument('--new-ip',
+    parser.add_argument('--host-ip',
                         metavar="<key=value>",
                         action='append',
-                        help='specify new ip address for current inventory, this argument '
-                             'need to use with key-value pairs combined with "=", the key '
-                             'must be the host name defined in the inventory yaml files, '
-                             'e.g. zuul01, nodepool02, allinone01.',
+                        help='override new ip address for current inventory, '
+                             'this argument need to use with key-value pairs '
+                             'combined with "=", the key must be the host name'
+                             ' defined in the inventory yaml files, e.g. '
+                             'zuul01, nodepool02, allinone01.',
                         )
     parser.add_argument('--with-vars',
                         action='store_true',
-                        help='show the hosts graph of ansible inventory with vars, specify with '
-                             'key-value pairs combined with "="',
-                        )
-    parser.add_argument('--graph',
-                        action='store_true',
-                        help='show the hosts graph of ansible inventory',
-                        )
-    parser.add_argument('--show-old-ips',
-                        action='store_true',
-                        help='show the old hosts ip address',
-                        )
-    parser.add_argument('--show-new-ips',
-                        action='store_true',
-                        help='show the new hosts ip address',
+                        help='show the hosts graph of ansible inventory with '
+                             'vars. It must be used together with --action '
+                             'show-graph.'
                         )
     return parser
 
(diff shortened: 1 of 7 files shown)


inventory/inventory.py
```diff
@@ -49,20 +49,19 @@ def main():
     args_group = parser.add_mutually_exclusive_group(required=True)
     args_group.add_argument('--list', action='store_true',
                             help='List inventories')
-    args_group.add_argument('--old-host-ips', action='store_true',
-                            help='List old host ips')
-    args_group.add_argument('--new-host-ips', action='store_true',
-                            help='List new host ips')
+    args_group.add_argument('--show-ip', action='store_true',
```
liusheng, line 52: this looks better

deploy.py
```diff
@@ -13,105 +13,86 @@ def add_cli_args():
                                  "openlab-ha"],
                         help='OpenLab deployment type',
                         )
+    parser.add_argument('--action',
+                        choices=["deploy", "new-slave", "switch-role",
```
liusheng, line 17: I am not sure if this looks better, let's see other's opinions

deploy.py
```diff
@@ -13,105 +13,86 @@ def add_cli_args():
...
+                                 "show-graph", "show-ip"],
+                        default="deploy",
+                        help="The action that labkeeper supports.\n"
+                             "'deploy': create a new OpenLab CI environment.\n"
+                             "'new_slave': create new slave nodes base on "
+                             "current master nodes.\n"
+                             "'switch_role': switch the master/slave role of "
```
liusheng: and also , this help message is not enough, e.g. the new_slave required `--new-ips`

deploy.py
```diff
@@ -13,105 +13,86 @@ def add_cli_args():
...
+    parser.add_argument('--host-ip',
                         metavar="<key=value>",
                         action='append',
-                        help='specify new ip address for current inventory, this argument '
-                             'need to use with key-value pairs combined with "=", the key '
-                             'must be the host name defined in the inventory yaml files, '
-                             'e.g. zuul01, nodepool02, allinone01.',
-                        )
-    parser.add_argument('--with-vars',
-                        action='store_true',
```
liusheng: why remove this ?

deploy.py
```diff
@@ -13,105 +13,86 @@ def add_cli_args():
                                  "openlab-ha"],
                         help='OpenLab deployment type',
                         )
+    parser.add_argument('--action',
+                        choices=["deploy", "new-slave", "switch-role",
+                                 "show-graph", "show-ip"],
+                        default="deploy",
+                        help="The action that labkeeper supports.\n"
+                             "'deploy': create a new OpenLab CI environment.\n"
+                             "'new_slave': create new slave nodes base on "
```
liusheng: all the actions choices are combined with `-`, but in help message you use `_`

deploy.py
```diff
@@ -13,105 +13,86 @@ def add_cli_args():
...
-        specified_ips = dict([(d.partition('=')[::2]) for d in parsed_args.new_ip])
+    if parsed_args.host_ip:
+        specified_ips = dict([(d.partition('=')[::2])
+                              for d in parsed_args.new_ip])
         for host, ip in specified_ips.items():
             os.environ['OL_%s_IP' % host.upper()] = ip
 
-    if parsed_args.new_slave:
-        if not parsed_args.new_ip:
-            sys.exit("ERROR: Must specify the new slave hosts IPs with --new-ip")
```
liusheng, line 82: please add this limit

deploy.py
```diff
@@ -13,105 +13,86 @@ def add_cli_args():
...
-    ol_env_msg = '\n'.join(['%s=%s' % (k, os.environ[k]) for k in os.environ if k.startswith('OL_')])
+    ol_env_msg = '\n'.join(['%s=%s' % (k, os.environ[k]) for k in os.environ
+                            if k.startswith('OL_')])
     print("OpenLab deployment ENV:\n%s" % ol_env_msg)
     print('Ansible command:\n%s' % ' '.join(cmd))
     print("*" * 100)
-
     subprocess.call(cmd)
 
+    if parsed_args.action == 'new-slave':
```
liusheng, line 101: how to run config-new-slave.yaml without new slave deployment ?

deploy.py
wangxiyuan: Merge all actions into one can make cli easy to use IMO. :)

deploy.py
wangxiyuan: Not sure it's useful, I can add it back if there is some use cases?

deploy.py
wangxiyuan: Done

deploy.py
wangxiyuan: I prefer not, IMO If new_ip is provided, use it. If not, just use the IPs in inventory file. All my change for this PR is to make the cli as clean as it could

deploy.py
wangxiyuan: `switch_role` action will run it.

deploy.py
liusheng: it is useful I think, if you want to confirm some inventory config options before run deployment, this can show all the hostvars based on current inventory

wangxiyuan: Done

deploy.py
liusheng: it is easy to forget to update the IPs in inventory files, in new slave situation. what we added this cli tool is to get rid of manually modification of the inventory files. so I think this is required.
I do not thinks the shorter help messages can make things better... and cleaner

deploy.py
liusheng: I do not think this is more easy to use than before, personal thoughts,  `--action` make things more secluded and have to run `help` firstly and then to know how to choose an action and what actions supported, so lets see others opinions

deploy.py
liusheng: OK, got

deploy.py
wangxiyuan: Before this PR, when I try to use the `deploy` script, I find that there is a lot of parameters, they must be used in some specified ways, otherwise the result is unexpected. For example,

what will happen if both `switch-role` and `new-slave` are provided? or like ``switch-role` and `--graph` are both provided? `--vars` must be used when `--graph` is provided,  but what if `switch-role` and `--vars` are provided at the same time?
 
This are different actions, but they can be used at the same time in one command, Isn't is strange? For a user of `deploy` script, Even after reading the help message, I still don't know what is the correct combinations for these parameters.  I just understand how to use it after read the source code.

BTW, doesn't reading help message is always the first step to use a cli?

deploy.py
wangxiyuan: 1. Update the inventory file should be the first step when first time using `labkeeper` IMO.
2. `new_slave` action is not always happen,(maybe 1-6 months or even 1 year). Once using it, I assume users will always download the newest `labkeeper`, Then I can't imaging that `update the inventory file` is easy to be forgotten. Users will use a file or a tool that is updated long time ago? Amazing.
3. Keep the same behavior with `deploy` action, when deploy, the `new_ip` is not required.

liusheng: 1. that is the old way, and that is why I add this deploy.py, to provide a way to avoid modify the inventory files, so it is not necessary now. I remember we have discussed this and think it is better to support this.
2. so you only need to update the repo. do not need to modify any things, and once a mistake happened running the `new_slave` without update the inventory, it will re-deploy the slave, that may cause bad effect. 
3. why ?

deploy.py
liusheng: just like openstack command tool, also have many parameter, and there is not a `--action` to specify the `start`, `stop`, `reboot`, `rebuild` ... such as actions. you do not need care about these arguments if you clearly know what you want to do. for example, you want to deploy an ha deployment, you only need to specify the type. yes, I also think the new-slave and swith-role is not clear definitions. better need to change. 
there is not a `--vars` parameters. only has `--with-vars` and `-extra-vars`, actually, these two and `--graph` is from Ansible itself support, only provided here for convenient check and extension. so I don't think it is strange. one parameter only used with another parameter is also an usual usage.

I remember I have mentioned about the parameter validation, I told these is not for common users, we are developers and this is for developers, so I didn't spent time to add validation about two actions specified in the same time case... so I can improve to make you happy if I have enough time and if you really care about these. we have readme and the help message.

sure if I clearly know about what I am going to do, I don't need to ready the help messages firstly, e.g. I just want to deploy an allinone for testing. I think the README is enough.

liusheng: Again, I am not strongly against this, every one has different personal habit, if others think OK, go ahead please.

deploy.py
wangxiyuan: I don't mean to remove `new_ip`, I mean we should keep `new_slave` the same behavior with `deploy`. If it's required, it should be both required for these actions. You said we should not rely on  inventory, why `deploy` action doesn't force require `new_ip`? Either make it required or make it optional is OK, but it should be the same for all related actions.

deploy.py
bzhaoopenstack: Er, how about go first for test, as the coming deadline? And I think once we test the most part of cases, then we will write a good doc for decribing the issue we faced, if all of us think there is still a issue here at that moment, we can start refactor there. Haha.. so let's merge?

deploy.py
bzhaoopenstack: I'm afraid I cant stop myself to click "merge" for coming deadline. ;-)

deploy.py
liusheng: that is not a good reason.. but.. all right, if you like. please go.

deploy.py
liusheng: @wangxiyuan actually, at first, I also thought like you and did't made that argument required, but after some testing operations,  I changed the `--new-ip` required because I think it should be necessary, if you run new-slave and forget updating IPs. that may be a serious, because it will re-install the slave, maybe it will cause some data missing. if you make the `--new-ip` required. it will compulsively operators and let him know it will deploy the new slave part in two new nodes specified by the `--new-ip`.

bzhaoopenstack (approved):
Let's merge first as no more feedback. If there are still some issues here, we can fix in the future.. Let's go..

PR #89, opened 2019-05-21
Add openlab switch role cli

wangxiyuan:
Add a new cli `cluster switch`. This is used by labkeeper tools.

Operator can use it to switch master-slave role by hand.


bzhaoopenstack (approved):
LGTM

PR #91, opened 2019-05-22
Fix gearman and rsyncd HA control

wangxiyuan:
gearman and rsyncd actually runs under systemd, it's called gearman-job-server and rsyncd

We should use systemd to control it.


bzhaoopenstack (approved):
lgtm

PR #92, opened 2019-05-23
Support configuration on zk

bzhaoopenstack:
This patch support configuration on zk except token things and repo.
1. Each loop will fetch the latest configuration from zk except ENV
first startup.
2. remove the old logs-backup domain from dns change.

### DIFF
--- a/etc/ha_healthchecker/ha_healthchecker.py.j2
+++ b/etc/ha_healthchecker/ha_healthchecker.py.j2
@@ -15,42 +15,29 @@ import requests
 import six
 
 
-logging.basicConfig(
-    filename="/etc/openlab/ha_healthchecker/ha_healthchecker.log",
-    filemode='a',
-    format='%(asctime)s,%(msecs)d %(name)s %(levelname)s %(message)s',
-    datefmt='%H:%M:%S',
-    level=logging.DEBUG)
-LOG = logging.getLogger(__name__)
-
-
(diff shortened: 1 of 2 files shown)


etc/ha_healthchecker/ha_healthchecker.py.j2
```diff
@@ -15,42 +15,29 @@ import requests
...
-    filemode='a',
-    format='%(asctime)s,%(msecs)d %(name)s %(levelname)s %(message)s',
-    datefmt='%H:%M:%S',
-    level=logging.DEBUG)
-LOG = logging.getLogger(__name__)
-
+# Simpledns provider
+DOMAIN_NAME = 'openlabtesting.org'
+TARGET_ORI_IP = "{{ ori_master_ip }}"
+TARGET_CHANGE_IP = "{{ slave_ip }}"
```
wangxiyuan, line 21: This two should not be hard coded, since the cluster can be changed. Let's fix it in the next PR and add related configuration to zk as well.

wangxiyuan (approved):

etc/ha_healthchecker/ha_healthchecker.py.j2
bzhaoopenstack: Yeah, I agree that all the original hard codes/parameters in this py file can be stored into zk.

PR #94, opened 2019-05-23
Refresh dns public ip

wangxiyuan:
When HA related action happen, the dns public ip stored in zk should be refreshed as well.

### DIFF
--- a/playbooks/conf-new-slave.yaml
+++ b/playbooks/conf-new-slave.yaml
@@ -60,3 +60,14 @@
         section: "zookeeper"
         option: "hosts"
         value: "{{ groups['zookeeper'] | map('extract', hostvars, ['ansible_host']) | join(', ') }}"
+
+- name: Run openlabcmd
+  become: yes
+  hosts: zuul-master
+  tasks:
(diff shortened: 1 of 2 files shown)


playbooks/conf-new-slave.yaml
```diff
@@ -60,3 +60,14 @@
...
         value: "{{ groups['zookeeper'] | map('extract', hostvars, ['ansible_host']) | join(', ') }}"
+
+- name: Run openlabcmd
+  become: yes
+  hosts: zuul-master
+  tasks:
+    - name: Run openlab ha config set cmd
+      shell: |
+        openlab ha config set dns_master_public_ip '{{ hostvars[groups["zuul-master"][0]].ansible_host }}'
+        openlab ha config set dns_slave_public_ip '{{ hostvars[groups["zuul-slave"][0]].ansible_host }}'
```
bzhaoopenstack, line 71: Correct, the healthchecker will only use the zk configuration in non-initialization cases.

bzhaoopenstack (approved):

PR #96, opened 2019-05-23
Add logrotate for healthchecker

bzhaoopenstack:
1. This patch split the log file of healthchecker with 5 part, the
limitation is 10M for the file size.
2. Allow set slave down state when can not pingable and heartbeat timeout. If we recover the slave node, leave a way to set itself status from down to up

### DIFF
--- a/etc/ha_healthchecker/ha_healthchecker.py.j2
+++ b/etc/ha_healthchecker/ha_healthchecker.py.j2
@@ -3,6 +3,7 @@ import configparser
 import datetime
 import json
 import logging
+from logging import handlers
 import socket
 import subprocess
 import time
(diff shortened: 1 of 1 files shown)


wangxiyuan (changes requested):
I got this error when test with this pr

ValueError: 'stream' or 'filename' should not be specified together with 'handlers'

wangxiyuan:
> I got this error when test with this pr
> 
> ValueError: 'stream' or 'filename' should not be specified together with 'handlers'

Ignore my comment. I forgot remove `filename` locally.

wangxiyuan (approved):

kiwik (approved):
Thank you.

PR #98, opened 2019-05-24
Fix some HA bug

wangxiyuan:
1. change `heartbeat_timeout_second` default value to 600 to make the HA switch check more reasonable.
2. When updating dns, the related config options should be updated as well.
3. When the node's status is down, healthchecker should works as refresher.


bzhaoopenstack (approved):
lgtm, we review face to face.

PR #103, opened 2019-05-25
Disable rsyncd during switch role

bzhaoopenstack:
For rsync,
We need to disable the rsyncd service on ori slave nodes.

### DIFF
--- a/playbooks/ha_log_cfg_sync.yaml
+++ b/playbooks/ha_log_cfg_sync.yaml
@@ -70,7 +70,7 @@
         set -xe
         systemctl disable rsyncd.service || true
       args:
-      executable: /bin/bash
+        executable: /bin/bash
 
 - name: Setup crontab job for sync zuul jobs logs files
   hosts: logserver-master
--- a/playbooks/ha_log_cfg_sync.yaml
+++ b/playbooks/ha_log_cfg_sync.yaml
@@ -65,6 +65,13 @@
         content: "{{ rsync_password }}"
         mode: 0600
 
+    - name: Teardown original slave rsyncd(when slave master switched)
+      shell: |
+        set -xe
+        systemctl disable rsyncd.service || true
+      args:
+      executable: /bin/bash
+
 - name: Setup crontab job for sync zuul jobs logs files
   hosts: logserver-master
   become: yes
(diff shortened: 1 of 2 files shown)


wangxiyuan (changes requested):
After f2f review, this should be moved into `ha_log_cfg_sync.yaml`

wangxiyuan (approved):

wangxiyuan (commented):
wait, seems L59 already did it?

bzhaoopenstack:
No , you just killed but not disable the service, when the host restart , the service will still be launched. We must disable it.

wangxiyuan:
> No , you just killed but not disable the service, when the host restart , the service will still be launched. We must disable it.

Make sense to me. Let's wait God Sheng's idea.

playbooks/ha_log_cfg_sync.yaml
```diff
@@ -65,6 +65,13 @@
         content: "{{ rsync_password }}"
         mode: 0600
 
+    - name: Teardown original slave rsyncd(when slave master switched)
+      shell: |
+        set -xe
+        systemctl disable rsyncd.service || true
+      args:
+      executable: /bin/bash
```
liusheng: need a indent at the head of L73(event this is ok, not a good habit :) )

playbooks/ha_log_cfg_sync.yaml
```diff
@@ -65,6 +65,13 @@
         content: "{{ rsync_password }}"
         mode: 0600
 
+    - name: Teardown original slave rsyncd(when slave master switched)
```
huangtianhua, line 68: Just a question: I can't find any changes you mentioned in comment msg: We also need to remove the related crontab on ori master nodes, which are already done in ha_log_cfg_sync playbook?

bzhaoopenstack: Yeah, the part you mentioned already done in that file. I will udpate the commit mesg

playbooks/ha_log_cfg_sync.yaml
bzhaoopenstack: oh, yes...

liusheng (approved):

PR #105, opened 2019-05-27
openlabcmd python3 only

wangxiyuan:
Since the main service like zuul or nodepool is python3 only and py2 is already deprecated, openlabcmd should use py3 as well.

Remove py2 support.

Related-Bug: theopenlab/openlab#272


kiwik (approved):
LGTM, thank you

PR #106, opened 2019-05-27
HA update

wangxiyuan:
1. Add zookeeper node HA check function
2. Remove some useless code.
3. Fixed a bug that unnecessary service won't be started at once when HA switch happen.


bzhaoopenstack (approved):
confirm after test. lgtm

PR #107, opened 2019-05-27
Add cirros-0.4.0/cirros-0.3.5 check

Yikun:
The cirros-0.4.0/0.3.5 are used by job, we enable the preparation
of these images is good for speeding up of job.

Part of https://github.com/theopenlab/openlab/issues/232

### DIFF
--- a/openlabcmd/openlabcmd/plugins/jobs/image.py
+++ b/openlabcmd/openlabcmd/plugins/jobs/image.py
@@ -1,6 +1,7 @@
 import subprocess
 
 from openlabcmd.plugins.base import Plugin
+from openlabcmd.plugins.recover import Recover
 
 
 class ImagePlugin(Plugin):
@@ -15,6 +16,9 @@ class ImagePlugin(Plugin):
         image_check = 'openstack --os-cloud %s image list ' \
                       '-c "Name" -c "Status" -f value | grep cirros' % self.cloud
         res = subprocess.getoutput(image_check)
-        if "cirros" not in res:
+        if "cirros-0.3.5" not in res:
             self.failed = True
-            self.reasons.append("- Image: cirros image not found.")
+            self.reasons.append(Recover.IMAGE_CIRROS_035)
+        if "cirros-0.4.0" not in res:
+            self.failed = True
+            self.reasons.append(Recover.IMAGE_CIRROS_040)
--- a/openlabcmd/openlabcmd/plugins/recover.py
+++ b/openlabcmd/openlabcmd/plugins/recover.py
@@ -9,6 +9,8 @@ class Recover(Enum):
     NETWORK = 5
     NETWORK_SUBNET = 6
     NETWORK_SUBNET_CIDR = 7
+    IMAGE_CIRROS_035 = 8
+    IMAGE_CIRROS_040 = 9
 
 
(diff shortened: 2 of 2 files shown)


wangxiyuan (approved):

openlabcmd/openlabcmd/plugins/jobs/image.py
```diff
@@ -15,6 +16,9 @@ def check(self):
         image_check = 'openstack --os-cloud %s image list ' \
                       '-c "Name" -c "Status" -f value | grep cirros' % self.cloud
         res = subprocess.getoutput(image_check)
-        if "cirros" not in res:
+        if "cirros-0.3.5" not in res:
             self.failed = True
-            self.reasons.append("- Image: cirros image not found.")
+            self.reasons.append(Recover.IMAGE_CIRROS_035)
+        if "cirros-0.4.0" not in res:
```
huangtianhua, line 22: First, if cirros-0.4.0 image in public cloud, we append the reason Recover.IMAGE_CIRROS_035, and if --recover specified, then we will download and register cirros 0.3.5 image for the cloud? Same case if there is only cirros-0.3.5 in cloud. Second, if there is no cirros image in cloud, we append two reasons, then if --recover specified, the cloud will download and register two cirros images, right? I don' t know whether it's ok, see the code like "_IMAGE=$(openstack image list | grep -i cirros | head -n 1)" not sure which one it will return and the test expected.

Yikun: I understand your worries, but I think if the commands with "head -n 1" means the job just want a image but don't need a specific version image, so it's not a problem.

But I only saw a problem line [1] in cirros search results[2]:

> src_img_name=$(openstack image list -f value -c Name | grep -Eo "cirros-.*" | grep -v ramdisk | grep -v kernel)

I think we need add a `head -n1` for it?

[1] https://github.com/theopenlab/openlab-zuul-jobs/blob/97b0fe4b/playbooks/packer-functional-devstack/run.yaml#L19
[2] https://github.com/theopenlab/openlab-zuul-jobs/search?p=1&q=cirros&unscoped_q=cirros

Yikun: And `openstack image show` has potential duplicate image problem(like `More than one resource exists with the name or ID 'cirros-0.3.5-x86_64-disk'.`)

So, maybe we should change all the image get script to `openstack image list | grep xxx | head -n 1`?

huangtianhua: Sorry, I am not sure, please check them. And not sure if we can use cirros-0.4.0 only?

liusheng:
Fow now, all the public clouds jobs have been disabled. the `cirrors` images existence check should only be in the `OTC`, `Telefonica`, `Orange`, `Huaweicloud`. So I am not sure if this is necessary. maybe just leave it in the jobs definitions ? and also. not all the public clouds support uploading cirrors image with `.img` format. I remember the Huaweicloud only support itself `vhd` format, your `recover` functionality may  not work in that case. also seems the disk size also has requirements. personal thoughts :)

Yikun:
@liu-sheng OK, I will close this pr, and I agree it's not a very necessary check in here.

PR #108, opened 2019-05-27
Add zookeeper_connect_timeout option

wangxiyuan:
Make zookeper client connection timeout configurable.

### DIFF
--- a/etc/openlabcmd/openlab.conf.j2
+++ b/etc/openlabcmd/openlab.conf.j2
@@ -3,3 +3,4 @@ cloud_conf = /etc/openstack/clouds.yaml
 
 [ha]
 zookeeper_hosts = {{ groups['zookeeper'] | map('extract', hostvars, 'ansible_host') | join(', ') }}
+zookeeper_connect_timeout = 5
--- a/openlabcmd/openlabcmd/zk.py
+++ b/openlabcmd/openlabcmd/zk.py
@@ -84,19 +84,32 @@ class ZooKeeper(object):
             return True
         return self.client.state == KazooState.LOST
 
-    def connect(self, hosts=None, read_only=False):
+    def connect(self, hosts=None, timeout=None, read_only=False):
         if not hosts:
             if not self.config:
                 raise exceptions.ClientError('Either config object or hosts '
                                              'string should be provided.')
-            else:
-                try:
-                    hosts = self.config.get('ha', 'zookeeper_hosts')
-                except configparser.NoOptionError:
-                    raise exceptions.ClientError(
-                        "The config doesn't [ha]zookeeper_hosts option.")
+            try:
+                hosts = hosts or self.config.get('ha', 'zookeeper_hosts')
+            except (configparser.NoOptionError, configparser.NoSectionError):
+                raise exceptions.ClientError(
+                    "The config doesn't contain [ha]zookeeper_hosts option.")
+
+        if not timeout:
+            timeout = self.config.get('ha', 'zookeeper_connect_timeout',
+                                      fallback=5)
+        try:
+            timeout = int(timeout)
+        except ValueError:
+            raise exceptions.ClientError("zookeeper_connect_timeout "
+                                         "should be int-like format.")
+        if timeout <= 0:
+            raise exceptions.ClientError("zookeeper_connect_timeout "
+                                         "should be larger than 0.")
+
         if self.client is None:
-            self.client = KazooClient(hosts=hosts, read_only=read_only)
+            self.client = KazooClient(hosts=hosts, timeout=timeout,
+                                      read_only=read_only)
             self.client.add_listener(self._connection_listener)
             # Manually retry initial connection attempt
             while True:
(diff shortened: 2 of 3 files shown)


openlabcmd/openlabcmd/zk.py
```diff
@@ -84,19 +84,28 @@ def lost(self):
...
+                    "The config doesn't contain [ha]zookeeper_hosts option.")
+            try:
+                timeout = timeout or self.config.get('ha', 'zookeeper_timeout')
+                timeout = int(timeout)
+            except configparser.NoOptionError:
+                raise exceptions.ClientError(
+                    "The config doesn't contain [ha]zookeeper_timeout option.")
+            except ValueError:
+                raise exceptions.ClientError("zookeeper_timeout should be "
+                                             "ini-like format.")
```
huangtianhua: int?

openlabcmd/openlabcmd/zk.py
```diff
@@ -84,19 +84,28 @@ def lost(self):
...
+                hosts = hosts or self.config.get('ha', 'zookeeper_hosts')
+            except configparser.NoOptionError:
+                raise exceptions.ClientError(
+                    "The config doesn't contain [ha]zookeeper_hosts option.")
+            try:
+                timeout = timeout or self.config.get('ha', 'zookeeper_timeout')
+                timeout = int(timeout)
+            except configparser.NoOptionError:
+                raise exceptions.ClientError(
+                    "The config doesn't contain [ha]zookeeper_timeout option.")
```
liusheng: why this is mandatory ？ does it has default value ?

etc/openlabcmd/openlab.conf.j2
```diff
@@ -3,3 +3,4 @@ cloud_conf = /etc/openstack/clouds.yaml
 
 [ha]
 zookeeper_hosts = {{ groups['zookeeper'] | map('extract', hostvars, 'ansible_host') | join(', ') }}
+zookeeper_timeout = 5
```
huangtianhua: Maybe named 'zookeeper_connect_timeout' is better.

openlabcmd/openlabcmd/zk.py
liusheng: if so, does the `config` object can define config options with type limit, required, and default value ?

openlabcmd/openlabcmd/zk.py
```diff
@@ -84,19 +84,29 @@ def lost(self):
...
+            except configparser.NoOptionError:
+                raise exceptions.ClientError(
+                    "The config doesn't contain [ha]zookeeper_hosts option.")
+            try:
+                timeout = timeout or self.config.get(
+                    'ha', 'zookeeper_connect_timeout')
+                timeout = int(timeout)
+            except configparser.NoOptionError:
+                raise exceptions.ClientError(
+                    "The config doesn't contain [ha]zookeeper_timeout option.")
```
huangtianhua: missed.

wangxiyuan: Done

openlabcmd/openlabcmd/zk.py
wangxiyuan: 1. Done, default is 5.
2. I didn't find a way to do it.

openlabcmd/openlabcmd/zk.py
```diff
@@ -84,19 +84,29 @@ def lost(self):
...
                                              'string should be provided.')
-            else:
-                try:
-                    hosts = self.config.get('ha', 'zookeeper_hosts')
-                except configparser.NoOptionError:
-                    raise exceptions.ClientError(
-                        "The config doesn't [ha]zookeeper_hosts option.")
+            try:
+                hosts = hosts or self.config.get('ha', 'zookeeper_hosts')
+            except configparser.NoOptionError:
```
Yikun: nit(It's not the problem in this change): Should we also add NoSectionError in here.
[1] https://github.com/python/cpython/blob/45d75fa/Lib/configparser.py#L782

openlabcmd/openlabcmd/zk.py
```diff
@@ -84,19 +84,29 @@ def lost(self):
...
+                raise exceptions.ClientError(
+                    "The config doesn't contain [ha]zookeeper_hosts option.")
+
+        if not timeout:
+            timeout = self.config.get('ha', 'zookeeper_connect_timeout',
+                                      fallback=5)
+        try:
+            timeout = int(timeout)
```
Yikun, line 102:
```python
>> int(-1)
-1
```
Maybe could add some limitation on positive value.

Yikun (changes requested):

bzhaoopenstack (approved):
lgtm

openlabcmd/openlabcmd/zk.py
bzhaoopenstack: I saw this is fixed  in the following line.

PR #110, opened 2019-05-28
Fix zookeeper configuration

bzhaoopenstack:
This patch fix the zk config when using Public IP.
We must configure the zk server to bind the local direct connect NIC.


wangxiyuan:
recheck

wangxiyuan (approved):

PR #112, opened 2019-05-28
Depoly real env found issues fix

bzhaoopenstack:
This PR fix list:
1. disable zuul debug level
2. Install openstackclient on all nodes for fitting checker requirements
3. Fix some nodes lacks setuptools before using pip install
4. Fix zookeeper local configuration to bind local interface ip.
5. Fix "create modsecurity config file" task in apache.yaml if we run 2nd time labkeeper, it will fail
6. Fix zuul-web exception to correct the binding IP in zuul.conf
7. Fix zuul-executor exception as lacking ansible 2.5 2.6 2.7
8. Fix zuul scheduler exception for correct zuul keys dir from 755 to 700
9. Fix nodepool build image failed as openstack-dev and openstack-infra

### DIFF
--- a/openlabcmd/requirements.txt
+++ b/openlabcmd/requirements.txt
@@ -1,6 +1,7 @@
 configparser
 enum;python_version=='2.7'
 kazoo
+openstackclient
 pbr>=1.3
 prettytable
 pytz
--- a/playbooks/prerequisites.yaml
+++ b/playbooks/prerequisites.yaml
@@ -27,7 +27,7 @@
       package:
         name: "{{ item }}"
       with_items:
-        - python-setuptools
+        - python3-setuptools
         - dbus
         - libffi-dev
         - libssl-dev
--- a/playbooks/prerequisites.yaml
+++ b/playbooks/prerequisites.yaml
@@ -37,6 +37,8 @@
         - python-dev
         - python3-pip
         - python-pip
+        - python-openstackclient
+        - python3-openstackclient
 
     - name: Install pip packages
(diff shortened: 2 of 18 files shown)


playbooks/prerequisites.yaml
```diff
@@ -27,6 +27,7 @@
       package:
         name: "{{ item }}"
       with_items:
+        - python-setuptools
```
wangxiyuan, line 30: python3-setuptools should be added as well.

wangxiyuan (changes requested):

huangtianhua (approved):

wangxiyuan:
Let's wait a moment in order to fix all related bug in one PR.

playbooks/prerequisites.yaml
```diff
@@ -27,6 +27,8 @@
...
+        - python3-setuptools
```
Yikun, line 31: and `openstack` commands?

openlabcmd/requirements.txt
```diff
@@ -1,6 +1,7 @@
 configparser
 enum;python_version=='2.7'
```
wangxiyuan, line 2: not related to this patch:  enum can be removed, since openlabcmd is only for py3 now.

Yikun:
recheck

wangxiyuan:
recheck

bzhaoopenstack:
recheck

wangxiyuan (approved):

PR #114, opened 2019-05-29
Fix the cron PATH in labcheck job

Yikun:
Cron jobs run in very minimal environment, the regular shell
setup never happens, so `/usr/local/bin` is not in the $PATH.

This patch add `/usr/local/bin` to PATH, which would be used
by `hub` and `openlab` cmd.

related: https://github.com/theopenlab/openlab/issues/232

### DIFF
--- a/labcheck/lab_check_report.sh
+++ b/labcheck/lab_check_report.sh
@@ -2,6 +2,9 @@ github_token=$1
 
 set -e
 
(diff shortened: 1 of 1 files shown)


huangtianhua (approved):

huangtianhua (approved):
LGTM, thanks.

kiwik (approved):
Thank you

PR #124, opened 2019-05-30
Support change github webapp

bzhaoopenstack:
This patch uses the original way to change github webapp by python crawler.
So we have to provider the github user password, as we can not use the
oauth token to simulate the human user login, also the github api doesnt
provide a way to change the app webhook. So python crawler is the only
to implement this.

1. Add 3 more option into local config, including github username, user
password, and app name.
2. Also add test and normal case for labkeeper to test, such as moo-ai


huangtianhua:
recheck

bzhaoopenstack:
recheck

bzhaoopenstack:
recheck

etc/ha_healthchecker/ha_healthchecker.py.j2
```diff
@@ -25,9 +27,15 @@ CONFIG_LOCAL = {
 {% if use_test_account %}
     'github_repo': "{{ test_repo_name }}",
     'github_user_token': "{{ test_github_token }}",
+    'github_user_passwd': "{{ test_github_password }}",
+    'github_user_name': "{{ test_github_user_name }}",
+    'github_app_name': "{{ test_github_app_name }}"
```
wangxiyuan, line 33: please add them into openlabcmd

wangxiyuan (changes requested):

etc/ha_healthchecker/ha_healthchecker.py.j2
```diff
@@ -25,9 +28,15 @@ CONFIG_LOCAL = {
...
```
wangxiyuan, line 33: what's the value of test_github_app_name? we may create one in moo?

etc/ha_healthchecker/ha_healthchecker.py.j2
```diff
@@ -780,23 +861,59 @@ class Switcher(Base):
...
+
+
     def _load(self, **kwargs):
         for attr, value in kwargs.items():
             if not hasattr(self, attr):
                 if not value:
                     value = CONFIG_LOCAL.get(attr)
+                else:
+                    if attr in ['github_user_passwd', 'dns_provider_token',
+                                'github_user_token']:
```
wangxiyuan: it's good set a variable for the list, like `base64_encoded_options` so that we can update it easily in the future just in one place.

etc/ha_healthchecker/ha_healthchecker.py.j2
```diff
@@ -25,9 +28,15 @@ CONFIG_LOCAL = {
...
     'github_repo': "{{ test_repo_name }}",
     'github_user_token': "{{ test_github_token }}",
+    'github_user_passwd': "{{ test_github_password }}",
+    'github_user_name': "{{ test_github_user_name }}",
+    'github_app_name': "{{ test_github_app_name }}"
 {% else %}
     'github_repo': 'theopenlab/openlab',
     'github_user_token': "{{ github_token }}",
+    'github_user_passwd': "{{ github_password }}",
+    'github_user_name': "theopenlab-ci",
```
huangtianhua: 1. Why not use {{ github_username}} here instead of hard code?
2. I don't think we have the var 'github_password' anymore, we use github_token instead of password, right?

etc/ha_healthchecker/ha_healthchecker.py.j2
```diff
@@ -25,9 +28,15 @@ CONFIG_LOCAL = {
 {% if use_test_account %}
     'github_repo': "{{ test_repo_name }}",
     'github_user_token': "{{ test_github_token }}",
+    'github_user_passwd': "{{ test_github_password }}",
```
huangtianhua, line 31: So here will be better to use github auth token instead?

bzhaoopenstack: There is no way to use oauth token to update the github app webhook, github api doesnt provider that. So we use python crawlers to simulate the normal human to login Github and change the webhook on a browser..

etc/ha_healthchecker/ha_healthchecker.py.j2
bzhaoopenstack: oh, I forget to add this in the inventory.

etc/ha_healthchecker/ha_healthchecker.py.j2
bzhaoopenstack: Yeah, I'm very clear what is done here, and I know very clear not use the password here.  But we could get only 1 way to change app webhook. Github API doesn't provide an API to change it[1].. So we only can simulate a human user to login github and change the app webhook by a python crawlers

reference:
[1] https://developer.github.com/v3/apps/

etc/ha_healthchecker/ha_healthchecker.py.j2
bzhaoopenstack: done

etc/ha_healthchecker/ha_healthchecker.py.j2
bzhaoopenstack: how about named "test-ha-app"?

wangxiyuan: ++

wangxiyuan (approved):

wangxiyuan:
Get it merge since maybe nobady knows what's is done here. lol

PR #130, opened 2019-06-03
Add --delete for log rsync

wangxiyuan:
Once the logs on master is cleaned up, the related logs on slave should be cleaned as well. Otherwise the logs on slave will be much and much.


bzhaoopenstack (approved):
gg

PR #134, opened 2019-06-04
Make ha_healthcheck as a service

wangxiyuan:
Change ha_healthchecker from script to a service in order to make it more official and easier controllable.

NOTE: all code is now in process.py. It need to be divided into more modules later.


bzhaoopenstack (approved):
ggg

PR #140, opened 2019-06-05
[HA]Hardcode log file location

wangxiyuan:
Hardcode log file location to '/var/log'. Drop the related config option.


bzhaoopenstack (approved):
lgtm

PR #141, opened 2019-06-05
add HA sg repair

bzhaoopenstack:
This PR add a ha check cli for checking existing node SG configuration.

summary:
openlab ha cluster repair --security-group
This cmd will try to check and repair the SG if find errors.

openlab ha cluster repair --security-group --dry-run
This cmd will only check the SG of existing nodes.

If we launch a new HA deployment, such as new-slave. We can just run this cmd(if the deployment cloud already contains SG 'openlab-ha-ports'), it will fix the all ha related SG rules automatically.

### DIFF
--- a/openlabcmd/openlabcmd/cli.py
+++ b/openlabcmd/openlabcmd/cli.py
@@ -187,6 +187,14 @@ class OpenLabCmd(object):
             'switch', help='Switch Master and Slave role.')
         cmd_ha_service_get.set_defaults(func=self.ha_cluster_switch)
 
+        # openlab ha cluster check
+        cmd_ha_cluster_check = cmd_ha_cluster_subparsers.add_parser(
+            'check', help='HA deployment check.')
+        cmd_ha_cluster_check.set_defaults(func=self.ha_cluster_check)
+        cmd_ha_cluster_check.add_argument(
+            '--dry-run', help='Check the Security Group of HA deployment.',
+            action='store_true')
+
     def _add_ha_config_cmd(self, parser):
         # openlab ha cluster
         cmd_ha_config = parser.add_parser('config',
@@ -391,6 +399,14 @@ class OpenLabCmd(object):
             print("Switch success")
         except exceptions.OpenLabCmdError:
             print("Switch failed")
+    
+    @_zk_wrapper
+    def ha_cluster_check(self):
+        try:
+            self.zk.check_deployment(is_dry_run=self.args.dry_run)
+            print("Check success")
+        except exceptions.OpenLabCmdError:
+            print("Check failed")
 
     @_zk_wrapper
     def ha_config_list(self):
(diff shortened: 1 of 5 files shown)


openlabcmd/openlabcmd/cli.py
```diff
@@ -187,6 +187,19 @@ def _add_ha_cluster_cmd(self, parser):
             'switch', help='Switch Master and Slave role.')
         cmd_ha_service_get.set_defaults(func=self.ha_cluster_switch)
 
+        # openlab ha cluster check
+        cmd_ha_cluster_check = cmd_ha_cluster_subparsers.add_parser(
+            'check', help='HA deployment check.')
```
wangxiyuan: So actually this command does `check` and `fix`, maybe the word `check` here is a little misleading. How about `openlab ha cluster repair`?

AND the  help message should contain more detail. at least it should contain what will be done?

openlabcmd/openlabcmd/cli.py
```diff
@@ -391,6 +404,16 @@ def ha_cluster_switch(self):
             print("Switch success")
         except exceptions.OpenLabCmdError:
             print("Switch failed")
+    
+    @_zk_wrapper
+    def ha_cluster_check(self):
+        # TODO(bz) This check may support other function
```
wangxiyuan: ++

openlabcmd/openlabcmd/zk.py
```diff
@@ -30,6 +32,17 @@
...
+RSYNCD_HA_PORT = 873
+ZOOKEEPER_HA_PORTS = [2181, 2888, 3888]
+MYSQL_HA_PORT = 3306
+
+HA_PORTS = []
+for ports in [RSYNCD_HA_PORT, ZOOKEEPER_HA_PORTS, MYSQL_HA_PORT]:
+    if isinstance(ports, list):
+        HA_PORTS.extend(ports)
+    else:
+        HA_PORTS.append(ports)
```
wangxiyuan: Please move L35-L44 to a new place, you can even create a new file call `repair.py`. Don't use global parameters by default.

openlabcmd/openlabcmd/cli.py
bzhaoopenstack: Yeah, I agree with u. This cmd can check and fix. Will add in the next commit

openlabcmd/openlabcmd/zk.py
bzhaoopenstack: Right, lets add a new file, even named a 'common.py' to store these values.

openlabcmd/openlabcmd/cli.py
```diff
@@ -187,6 +187,19 @@ def _add_ha_cluster_cmd(self, parser):
             'switch', help='Switch Master and Slave role.')
         cmd_ha_service_get.set_defaults(func=self.ha_cluster_switch)
 
+        # openlab ha cluster repair
+        cmd_ha_cluster_check = cmd_ha_cluster_subparsers.add_parser(
```
wangxiyuan: `cmd_ha_cluster_check` should be changed as well.

wangxiyuan (commented):
a nit

wangxiyuan (approved):

PR #144, opened 2019-06-06
Add configurable service restart count

bzhaoopenstack:
1. Add new attr 'restarted_count' into service obj
2. Add new option 'service_count_allow_to_raise' into cluster
configuration, default is 3.

Then the restart service operation will delay for 3 times, if the last
status is still down, the logic will set the service status down in zk.

### DIFF
--- a/ha_healthchecker/ha_healthchecker/action/refresher.py
+++ b/ha_healthchecker/ha_healthchecker/action/refresher.py
@@ -18,10 +18,13 @@ class Refresher(base.Action):
         cur_status = self._get_service_status(service_obj.name)
         update_dict = {}
         if cur_status == 'up':
-            if service_obj.status != 'up':
+            if (service_obj.status != 'up' and
+                    service_obj.restarted_count <
+                    self.cluster_config.service_count_allow_to_raise):
                 update_dict['status'] = 'up'
                 update_dict['restarted'] = False
                 update_dict['alarmed'] = False
+                update_dict['restarted_account'] = 0
                 self.LOG.debug("Fix Service %(name)s status from %(orig)s to "
                                "UP.", {'name': service_obj.name,
                                        'orig': service_obj.status})
@@ -32,9 +35,18 @@ class Refresher(base.Action):
                 self.LOG.debug("Service %(name)s is Restarting.",
                                {'name': service_obj.name})
             else:
-                update_dict['status'] = 'down'
-                self.LOG.debug("Service %(name)s is Down.",
-                               {'name': service_obj.name})
+                if (service_obj.restarted_count >
+                        self.cluster_config.service_count_allow_to_raise):
+                    update_dict['status'] = 'down'
+                    self.LOG.debug("Service %(name)s is Down.",
+                                   {'name': service_obj.name})
+                else:
+                    update_dict[
+                        'restarted_account'] = service_obj.restarted_count + 1
+                    self.LOG.debug("Service %(name)s continue in Restarting, "
+                                   "tried %(count)s times",
+                                   {'name': service_obj.name,
+                                    'count': service_obj.restarted_count})
         if update_dict:
             self.zk.update_service(service_obj.name, node_obj.name,
                                    **update_dict)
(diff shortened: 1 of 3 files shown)


wangxiyuan:
If we can retry restart the service 3 times in one loop, then we don't need a new property in zk. How do you think?

bzhaoopenstack:
Correct, if we use the original thought to design this, this PR is enough. But I think if we need to hold on the time between we restart the service. So the question is if we set it into 1 loop, the loop will hang by waiting the time between restart..

ha_healthchecker/ha_healthchecker/action/refresher.py
```diff
@@ -18,10 +18,13 @@ def _refresh_service(self, service_obj, node_obj):
         cur_status = self._get_service_status(service_obj.name)
         update_dict = {}
         if cur_status == 'up':
-            if service_obj.status != 'up':
+            if (service_obj.status != 'up' and
+                    service_obj.restarted_count <
+                    self.cluster_config.service_count_allow_to_raise):
```
wangxiyuan: I think no matter what's the restarted_count  value, once the service is up, it should be set to 0, so just `if (service_obj.status != 'up'` is enough.

ha_healthchecker/ha_healthchecker/action/refresher.py
```diff
@@ -32,9 +35,18 @@ def _refresh_service(self, service_obj, node_obj):
                 self.LOG.debug("Service %(name)s is Restarting.",
                                {'name': service_obj.name})
             else:
-                update_dict['status'] = 'down'
-                self.LOG.debug("Service %(name)s is Down.",
-                               {'name': service_obj.name})
+                if (service_obj.restarted_count >
```
wangxiyuan: `>=`, right?

ha_healthchecker/ha_healthchecker/action/refresher.py
```diff
@@ -32,9 +35,18 @@ def _refresh_service(self, service_obj, node_obj):
...
+                        self.cluster_config.service_count_allow_to_raise):
+                    update_dict['status'] = 'down'
+                    self.LOG.debug("Service %(name)s is Down.",
+                                   {'name': service_obj.name})
+                else:
+                    update_dict[
+                        'restarted_account'] = service_obj.restarted_count + 1
+                    self.LOG.debug("Service %(name)s continue in Restarting, "
```
wangxiyuan: `restarting`

openlabcmd/openlabcmd/zk.py
```diff
@@ -29,6 +29,7 @@
     'github_user_token': None,
     'heartbeat_timeout_second': 600,
     'logging_level': 'DEBUG',
+    'service_count_allow_to_raise': 3,
```
wangxiyuan: `service_restart_max_allow_times` is better.

wangxiyuan (changes requested):

openlabcmd/openlabcmd/zk.py
wangxiyuan: or just `service_restart_max_times`

ha_healthchecker/ha_healthchecker/action/refresher.py
bzhaoopenstack: Oh, yes.

ha_healthchecker/ha_healthchecker/action/refresher.py
bzhaoopenstack: done

ha_healthchecker/ha_healthchecker/action/refresher.py
bzhaoopenstack: done

openlabcmd/openlabcmd/zk.py
bzhaoopenstack: done

wangxiyuan (approved):

PR #153, opened 2019-06-13
Give zuul job more space

wangxiyuan:
The source code of some projects is not small. Like percona-server, it's more than 800M, so its job always hits `DISK_FULL` error in zuul.

This pr make the limit larger to ensure the big project can be ran as well.


bzhaoopenstack (approved):
lgtm

PR #165, opened 2019-07-02
Update log sync time

wangxiyuan:
Since sync the log from master to slave will cost much cpu and disk IO, so let's move the job to night.


bzhaoopenstack (approved):
lgtm

PR #166, opened 2019-07-02
Add orphan check for server, volume and fip

bzhaoopenstack:
Add a new plugin into labchecker for orphan check.
This plugin's ptype is also 'nodepool', as the source data comes from nodepool

### DIFF
--- a/openlabcmd/openlabcmd/cli.py
+++ b/openlabcmd/openlabcmd/cli.py
@@ -315,7 +315,7 @@ class OpenLabCmd(object):
             self._header_print(header)
             for plugin in plugins:
                 plugin.cloud = c
-                plugin.check_begin()
+                plugin = plugin.check_begin()
                 plugin.check()
                 plugin.check_end()
                 # the failed flag would be record when do check()
--- a/openlabcmd/openlabcmd/plugins/base.py
+++ b/openlabcmd/openlabcmd/plugins/base.py
@@ -26,16 +26,8 @@ class PluginMount(type):
     def register_plugin(cls, plugin):
         """Add the plugin to the plugin list and perform any registration logic"""
 
-        # create a plugin instance and store it
-        # optionally you could just store the plugin class and lazily instantiate
-        instance = plugin()
-
         # save the plugin reference
-        cls.plugins.append(instance)
-
-        # apply plugin logic - in this case connect the plugin to blinker signals
-        # this must be defined in the derived class
-        instance.register_signals()
+        cls.plugins.append(plugin)
 
 
 @six.add_metaclass(PluginMount)
@@ -45,16 +37,19 @@ class Plugin(object):
     ptype = 'Base'
     name = 'Base Check'
     failed = False
-    reasons = []
     nocolor = False
     experimental = False
+
+    def __init__(self):
+        self.reasons = []
 
     def register_signals(self):
         # print("%s has been loaded." % self.__class__.__name__)
         pass
 
-    def check_begin(self):
-        pass
+    @classmethod
+    def check_begin(cls):
+        return cls()
 
     def check(self):
         pass
--- a/openlabcmd/openlabcmd/plugins/nodepool/orphanresource.py
+++ b/openlabcmd/openlabcmd/plugins/nodepool/orphanresource.py
@@ -1,10 +1,17 @@
-import copy
+import datetime
+import iso8601
+import json
+import requests
 import subprocess
+import yaml
 
 from openlabcmd.plugins.base import Plugin
 
 SERVER_WHITE_LIST = [
     '%s-openlab-zuul', '%s-openlab-nodepool', '%s-openlab-zookeeper']
+OPENSTACK_CLOUD_CFG_PATH = '/etc/openstack/clouds.yaml'
+# DAYs
+RESOURCE_TIMEOUT = 1
 
 
 class OrphanResourcePlugin(Plugin):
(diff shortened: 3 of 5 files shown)


openlabcmd/openlabcmd/plugins/nodepool/orphanresource.py
```diff
@@ -0,0 +1,120 @@
...
+class OrphanResourcePlugin(Plugin):
+
+    # This Plugin is only for checking, as it's a dangerous action for each.
+    ptype = 'nodepool'
+    name = 'OrphanResource'
+
+    def check(self):
+        self.failed = False
+        if self.reasons:
+            self.reasons.pop()
```
wangxiyuan: bug for `self.reason`

openlabcmd/openlabcmd/plugins/nodepool/orphanresource.py
```diff
@@ -0,0 +1,120 @@
...
+        for line in res.split('\n'):
+            elements = line.split(' ')
+            if elements[1] not in mapping:
+                mapping[elements[1]] = {
+                    'nodepool_compute_mapping':
+                        [{'nodepool_id': elements[0],
+                          'compute_id': elements[2],
+                          'floatingip': elements[3]}],
+                    'compute_ids': [elements[2]],
+                    'floatingips': [elements[3]]}
```
wangxiyuan, line 186: seems only `compute_id` is used below, what are other ids used for?

wangxiyuan (changes requested):

openlabcmd/openlabcmd/plugins/nodepool/orphanresource.py
```diff
@@ -0,0 +1,120 @@
...
+            "| sort -t ' ' -k2")
+        ret, res = subprocess.getstatusoutput(get_nodepool_map_cmd)
+
+        mapping = {}
+        for line in res.split('\n'):
+            elements = line.split(' ')
+            if elements[1] not in mapping:
+                mapping[elements[1]] = {
+                    'nodepool_compute_mapping':
+                        [{'nodepool_id': elements[0],
```
ZhengZhenyu: This seems never used? and the name is somehow missleading, could we just don't set this?

openlabcmd/openlabcmd/plugins/nodepool/orphanresource.py
```diff
@@ -0,0 +1,120 @@
...
+        if self.reasons:
+            self.reasons.pop()
+        server_white_list = [elemnt % self.cloud
+                             for elemnt in copy.deepcopy(SERVER_WHITE_LIST)]
+
+        # For floatingip, we will check whether a fip didn't associated with a
+        # neutron port.
+        get_fip_cmd = ("openstack --os-cloud %s floating ip list "
+                       "-c 'ID' -c 'Floating IP Address' -c 'Port' | "
+                       "egrep '[a-z0-9]{8}-[a-z0-9]{4}-[a-z0-9]{4}-"
```
liusheng: maybe such as commands should be more simple ?
`openstack floating ip list -c 'Floating IP Address' -c 'Fixed IP Address' -f value |grep None| awk '{ print $1 }'`

openlabcmd/openlabcmd/plugins/nodepool/orphanresource.py
```diff
@@ -0,0 +1,120 @@
...
+                       "[a-z0-9]{4}-[a-z0-9]{12}' | awk '{print $2, $4, $6}' "
+                       "| grep None | awk '{print $1}'")
+        # For volume, we just check the available status volumes, as we cannot
+        # know whether a in-use volume is used for any purpose.
+        get_volume_cmd = (
+            "openstack --os-cloud %s volume list -c 'ID' -c 'Status' | "
+            "egrep '[a-z0-9]{8}-[a-z0-9]{4}-[a-z0-9]{4}-[a-z0-9]{4}-"
```
liusheng: `openstack volume list --status available -f value -c ID`

openlabcmd/openlabcmd/plugins/nodepool/orphanresource.py
```diff
@@ -0,0 +1,120 @@
...
+            "[a-z0-9]{12}' | grep available | awk '{print $2}'")
+        # For servers, we will get all servers for a provider, and skip the
+        # online openlab env servers.
+        get_server_cmd = (
+            "openstack --os-cloud %s server list -c 'ID' -c 'Name' | "
+            "egrep '[a-z0-9]{8}-[a-z0-9]{4}-[a-z0-9]{4}-[a-z0-9]{4}-"
```
liusheng: similar to above

openlabcmd/openlabcmd/plugins/nodepool/orphanresource.py
liusheng: and I want to remember, we may shouldn't treat all the volumes in `available` status as orphan volume. for some of the `cloud-provider-openstack` jobs, it will try to create volumes in cloud as PVs, when running jobs, the jobs may create volumes, and before attaching, the volume maybe in `available` status

openlabcmd/openlabcmd/plugins/nodepool/orphanresource.py
bzhaoopenstack: Yeah, it works. Thanks.

openlabcmd/openlabcmd/plugins/nodepool/orphanresource.py
bzhaoopenstack: So, if fip is in the same case, the fip not attach a real port during test. How about add another filter? Check the created_at.

openlabcmd/openlabcmd/plugins/nodepool/orphanresource.py
bzhaoopenstack: Nice.

openlabcmd/openlabcmd/plugins/nodepool/orphanresource.py
bzhaoopenstack: correct, I will remove this.

openlabcmd/openlabcmd/plugins/nodepool/orphanresource.py
bzhaoopenstack: Yeah, only compute_ids and floatingips are used. I will remove nodepool_compute_mapping

openlabcmd/openlabcmd/plugins/nodepool/orphanresource.py
liusheng: yes, so maybe what we can do is only to audit the orphan resource,  and then let operators to decide to delete or not..

openlabcmd/openlabcmd/plugins/base.py
```diff
@@ -45,16 +37,19 @@ class Plugin(object):
...
+    def __init__(self):
+        self.reasons = []
+
     def register_signals(self):
         # print("%s has been loaded." % self.__class__.__name__)
         pass
 
-    def check_begin(self):
-        pass
+    @classmethod
```
Yikun, line 50: -1, As I mentioned before, it's better to use ``__init__`` to create new plugin instance.

openlabcmd/openlabcmd/cli.py
```diff
@@ -315,7 +315,7 @@ def check(self):
             self._header_print(header)
             for plugin in plugins:
```
Yikun, line 316: -1, I prefer below style to run plugin:
```python
for p in plugins:
    # p is the plugin class, plugin is the plugin instance
    plugin = p() 
    plugin.check_begin()
    plugin.check()
    plugin.check_end()
```
And ``check_begin`` is designed for some reserve operation for pre-check(such as, import env, create some fake data, or something), so we should use the original ``__init__`` to initialize the class.

openlabcmd/openlabcmd/plugins/base.py
```diff
@@ -26,16 +26,8 @@ def __init__(cls, name, bases, attrs):
...
-        # optionally you could just store the plugin class and lazily instantiate
-        instance = plugin()
-
         # save the plugin reference
-        cls.plugins.append(instance)
-
-        # apply plugin logic - in this case connect the plugin to blinker signals
-        # this must be defined in the derived class
-        instance.register_signals()
+        cls.plugins.append(plugin)
```
Yikun, line 30: nit, So, now the plugin is the plugin class, I guess we could change ``plugins`` to ``Plugins`` to avoid some confusion? and we can also add some note on it.

openlabcmd/openlabcmd/plugins/base.py
```diff
@@ -45,16 +37,19 @@ class Plugin(object):
     ptype = 'Base'
```
Yikun, line 37: nit: Looks like all the member could be moved into ``__init__``

openlabcmd/openlabcmd/plugins/nodepool/orphanresource.py
```diff
@@ -0,0 +1,253 @@
...
+RESOURCE_TIMEOUT = 1
+
+
+class OrphanResourcePlugin(Plugin):
+
+    # This Plugin is only for checking, as it's a dangerous action for each.
+    ptype = 'nodepool'
+    name = 'OrphanResource'
+
+    def _get_endpoint_and_token(self):
```
Yikun, line 23: question: Not sure why we could not use the cli to get the token and endpoint?

Yikun (changes requested):

openlabcmd/openlabcmd/cli.py
wangxiyuan: +1

openlabcmd/openlabcmd/plugins/nodepool/orphanresource.py
wangxiyuan: use openstacksdk directly is easier and simpler.

openlabcmd/openlabcmd/plugins/nodepool/orphanresource.py
wangxiyuan: where is `floatingips` used.

Yikun:
FYI, the recent results [1] show that many manually created vm is found by this plugin, so it would be good add some white list mechanism into plugin.

[1] https://github.com/theopenlab/openlab/issues/314

wangxiyuan:
see: https://github.com/theopenlab/labkeeper/pull/171

PR #171, opened 2019-07-10
Add orphan check for server, volume and fip

wangxiyuan:
Add a new plugin into labchecker for orphan check.

### DIFF
--- a/openlabcmd/openlabcmd/plugins/jobs/image.py
+++ b/openlabcmd/openlabcmd/plugins/jobs/image.py
@@ -4,9 +4,11 @@ from openlabcmd.plugins.base import Plugin
 
 
 class ImagePlugin(Plugin):
-
     ptype = 'jobs'
     name = 'image'
+
+    def __init__(self, cloud, config):
+        super(ImagePlugin, self).__init__(cloud, config)
 
     def check(self):
         self.failed = False
--- a/openlabcmd/openlabcmd/plugins/nodepool/orphanresource.py
+++ b/openlabcmd/openlabcmd/plugins/nodepool/orphanresource.py
@@ -1,0 +1,128 @@
+import datetime
+import iso8601
+import subprocess
+
+import openstack
+
+from openlabcmd.plugins.base import Plugin
+
+
+class OrphanResourcePlugin(Plugin):
+    ptype = 'nodepool'
+    name = 'orphan_resource'
+
+    # This Plugin is only for checking, as it's a dangerous action for each.
+    def __init__(self, cloud, config):
+        super(OrphanResourcePlugin, self).__init__(cloud, config)
+        self.openstack_client = openstack.connect(cloud)
+        vm_common_white_list = [element % self.cloud
+                                for element in ['%s-openlab-zuul',
+                                                '%s-openlab-nodepool',
+                                                '%s-openlab-zookeeper']]
+        vm_addition_white_list = self.config.get(
+            'check', 'vm_white_list', fallback='').replace(' ', '').split(',')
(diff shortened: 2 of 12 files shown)


openlabcmd/openlabcmd/plugins/jobs/image.py
```diff
@@ -4,10 +4,12 @@
...
 class ImagePlugin(Plugin):
-
     ptype = 'jobs'
     name = 'image'
 
+    def __init__(self, cloud, config):
+        super(ImagePlugin, self).__init__(cloud, config)
+
     def check(self):
         self.failed = False
```
Yikun, line 14: we already set the `failed` and `reasons` in parent class, we can remove these now.

openlabcmd/openlabcmd/plugins/nodepool/orphanresource.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+class OrphanResourcePlugin(Plugin):
+    ptype = 'nodepool'
+    name = 'orphan_resource'
+
+    # This Plugin is only for checking, as it's a dangerous action for each.
+    def __init__(self, cloud, config):
+        super(OrphanResourcePlugin, self).__init__(cloud, config)
+        self.openstack_client = openstack.connect(cloud)
```
Yikun, line 17: In future we can move the openstack_client into a separate class, and the plugin is using the openstacksdk will be defined as:
```
class OrphanResourcePlugin(Plugin, OpenstackSDKHelper):
```

Yikun (approved):

openlabcmd/openlabcmd/plugins/nodepool/orphanresource.py
Yikun: I mean if we have lots of other methods use sdk to do some complex thing, but now we can keep it.

openlabcmd/openlabcmd/plugins/jobs/image.py
wangxiyuan: It's more like a `tag reset button`.

openlabcmd/openlabcmd/plugins/nodepool/orphanresource.py
wangxiyuan: Good point, it can be done once more plugin using openstacksdk

openlabcmd/openlabcmd/plugins/nodepool/orphanresource.py
```diff
@@ -0,0 +1,130 @@
...
+        self.vm_white_list = list(set(vm_common_white_list +
+                                      vm_addition_white_list))
+        self.volume_white_list = self.config.get(
+            'check', 'volume_white_list',
+            fallback='').replace(' ', '').split(',')
+
+        self.fip_white_list = self.config.get(
+            'check', 'fip_white_list', fallback='').replace(' ', '').split(',')
+
+    def _is_check_heart_beat_overtime(self, timestamp):
```
bzhaoopenstack: This function is meaningless. How about "_is_overtime" or something like this? This is not related with heartbeat.

openlabcmd/openlabcmd/plugins/nodepool/orphanresource.py
```diff
@@ -0,0 +1,130 @@
...
+        if not timestamp:
+            return True
+        timeout = RESOURCE_TIMEOUT
```
bzhaoopenstack: I saw the whitelist is moved to cfg. How about to do the same with timeout?

bzhaoopenstack (approved):
lgtm

PR #179, opened 2019-07-31
Add Router check into network plugin for labchecker

bzhaoopenstack:
Add new Router check into network plugin for labchecker.


openlabcmd/openlabcmd/plugins/nodepool/router.py
```diff
@@ -0,0 +1,67 @@
+import subprocess
```
wangxiyuan: I think router belongs to network. No need to create a new py file, just merge it inot network.py

openlabcmd/openlabcmd/plugins/constants.py
```diff
@@ -0,0 +1,4 @@
+from openlabcmd.plugins.recover import Recover
+
+# Need special process cases
+SPECIAL_CASES = [Recover.ROUTER_EXTERNAL_GW]
```
wangxiyuan: no need for this file, just merge thme into recover.py

openlabcmd/openlabcmd/plugins/base.py
```diff
@@ -91,7 +93,12 @@ def recover(self):
         print("Recover:")
         for r_code in self.reasons:
             if r_code in RECOVER_MAPS:
-                recover_cmd = RECOVER_MAPS[r_code]['recover'] % self.cloud
+                if (r_code in constants.SPECIAL_CASES and
+                        self.special_case_codes):
+                    recover_cmd = self.special_process(
```
Yikun: The `special_process` looks like a workaround, but not the final solution, this process is trying to solve the arg format of recover cmd, can we add the `arguments` to make the thing general?

    Recover.ROUTER_EXTERNAL_GW: {
        "recover": "openstack --os-cloud {} openstack router set openlab-router --external-gateway {} ",
        "reason": "- Router external gateway: openlab-subnet doesn't attach "
                  "on openlab-router.",
        "args": ["1", "2", "3"]
    }

When we need to format the recover, we can do something like
```python
"openstack --os-cloud {} openstack router set openlab-router --external-gateway {}".format(*["1", "2", "3"])
```

See more https://docs.python.org/3.6/library/string.html#format-examples

openlabcmd/openlabcmd/plugins/recover.py
```diff
@@ -45,4 +48,21 @@ class Recover(Enum):
                    "--network openlab-net --subnet-range=192.168.0.0/24",
         "reason": "- Subnet cidr: 192.168.0.0/24 is not found.",
     },
+    Recover.ROUTER: {
```
Yikun, line 58: As mentioned before, does the router depends on the subnet or network? if yes, maybe we should append the router in network plugin, or introduce some priority mechanism in labchecker in future?

Yikun (changes requested):

openlabcmd/openlabcmd/plugins/recover.py
bzhaoopenstack: Router creation doesn't depend on those resources, but our process flow does. So I move the code into network plugin.

openlabcmd/openlabcmd/plugins/nodepool/network.py
```diff
@@ -38,3 +38,49 @@ def check(self):
...
+
+        ext_net_id = res
+        router_gw = 'openstack --os-cloud %s router show ' \
+                    'openlab-router | grep %s' % (self.cloud, res)
+        ret, res = subprocess.getstatusoutput(router_gw)
+        if ret != 0:
+            self.failed = True
+            self.reasons.append(Recover.ROUTER_EXTERNAL_GW)
+            Recover.RECOVER_MAPS[Recover.ROUTER_EXTERNAL_GW][
+                'recover_args'].append(ext_net_id)
```
wangxiyuan: Do not modify global varable. It's dangerous and uncontrollable.

wangxiyuan (changes requested):

wangxiyuan (approved):

wangxiyuan:
LGTM

Yikun (approved):

PR #181, opened 2019-08-07
Add the prefix match on orphanresource plugin

Yikun:
Add the prefix match on orphanresource plugin, if server name starts with white name will be filtered.

### DIFF
--- a/openlabcmd/openlabcmd/plugins/nodepool/orphanresource.py
+++ b/openlabcmd/openlabcmd/plugins/nodepool/orphanresource.py
@@ -63,9 +63,17 @@ class OrphanResourcePlugin(Plugin):
             return
 
         servers = self.openstack_client.list_servers()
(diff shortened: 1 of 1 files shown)


openlabcmd/openlabcmd/plugins/nodepool/orphanresource.py
```diff
@@ -63,9 +63,17 @@ def check(self):
...
-                        if s['name'] not in self.vm_white_list]
-                        if servers else [])
+
+        real_servers = []
+        for s in servers:
+            skip = False
+            # if the vm name startswith the name in white list, will skip.
+            for w in self.vm_white_list:
+                if s['name'].startswith(w):
+                    skip = True
```
wangxiyuan, line 73: break

wangxiyuan (changes requested):

bzhaoopenstack (approved):
go

wangxiyuan (approved):

PR #182, opened 2019-08-07
Add retry limit for zookeeper connection

wangxiyuan:
Do not retry forever for zookeeper connection. Add the limit,
the default value is 5.


bzhaoopenstack (approved):
go

PR #189, opened 2019-08-19
 Refactor Router check

bzhaoopenstack:
We check the ext-net first, if ext-net failed, we must contact the cloud
administrator to see, if no ext-net, create the router is useless for
us.
Then if cannot find a router, we will recover for creating a router,
interface attachment and gateway setting.
if find a router, but didn't find it had been attached into a specific
network, we will attach it into openlab-subnet, also if ext-net is not
setted on router, we will add it into recover process.

### DIFF
--- a/openlabcmd/openlabcmd/plugins/nodepool/network.py
+++ b/openlabcmd/openlabcmd/plugins/nodepool/network.py
@@ -78,7 +78,11 @@ class NetworkPlugin(Plugin):
         if ret != 0:
             self.failed = True
             self.reasons.append(Recover.ROUTER_SUBNET_INTERFACE)
-            if ext_net_id not in res:
+            router_ext_check = 'openstack --os-cloud %s router show ' \
+                               'openlab-router | grep %s' % (
+                                self.cloud, ext_net_id)
+            re_ret, re_res = subprocess.getstatusoutput(router_ext_check)
+            if re_ret != 0:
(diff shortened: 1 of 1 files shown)


openlabcmd/openlabcmd/plugins/nodepool/network.py
```diff
@@ -39,15 +39,29 @@ def check(self):
...
+        ret, res = subprocess.getstatusoutput(router)
         if ret != 0:
             self.failed = True
-            if "More than one Openlab router exists" in res:
+            if "More than one Router exists" in res:
                 self.reasons.append(res)
             else:
                 self.reasons.append(Recover.ROUTER)
                 self.reasons.append(Recover.ROUTER_SUBNET_INTERFACE)
+                self.reasons.append(Recover.ROUTER_EXTERNAL_GW)
```
Yikun, line 62: L25 and L40 will return directly, and the router check will be skiped, that means in an empty env, we should do openlab check twice, does this meet your mind?

bzhaoopenstack: I know, but network and router resources have no relationship. So I need to bring nested check into every step?? When we do recover process, It checks the network is not exist, so we set Recover.Netowrk into Recovering process, but we don't know if Router exists, right? we should cover these cases.

PR #190, opened 2019-08-20
Clean up zk client if connect failed

wangxiyuan:
zk client obj should be cleaned up if connect failed so that it can be reconnected in the next loop.


bzhaoopenstack (approved):
LGTM

